PR #14, opened 2021-11-13
Refactor examples & add Getting Started Guide

ionicsolutions:


ionicsolutions:
@mw-vishnug

Two questions:

1.  Why does the Makefile need specific data? And what is the purpose of the Makefile? Is it actually building the project or merely triggering the tests?
2. What exactly is the purpose of the Spark demo from your point of view? Should this be an "Example" as well? Is this documented somewhere? Would it be an option to move it to `/demo` instead of `/docker_build` as it does not seem to actually build an `anovos` Docker container that can run arbitrary `anovos` code?

ionicsolutions:
I have completed all sections of the guide. Right now, there are only two open issues:

- [ ] We need to add the link to the page in the documentation (or elsewhere) that describes which versions of Spark/Java/Python the library is compatible with and how the `SparkSession` needs to be configured
- [ ] It doesn't seem to be possible to open the `datapane` report within the Jupyter environment due to "unknown origin" issues (`Blocking request from unknown origin`). Any ideas?

mw-vishnug:
> @mw-vishnug
> 
> Two questions:
> 
> 1. Why does the Makefile need specific data? And what is the purpose of the Makefile? Is it actually building the project or merely triggering the tests?

Agree that the Makefile should be generic and not for any specific data. 
Anovos was written to run as a pyspark application and therefore the current Makefile on ` build` in `make clean build` builds the "anovos.zip" package to be included while doing spark-submit. Other instructions set is to copy the `src/main/main.py` and the `config/configs.yaml` to the `dist/` folder, which has the configuration to be modified by the user for the user dataset. 
Additionally, currently it also copies the `data/income_dataset` as a sort of "default dataset" to do the standalone run in local environment. This part I think can be removed and the default `config/configs.yaml` file can be modified to point to the data from the new path `../example/data` so Makefile does not have to be tied to any dataset.

> 2. What exactly is the purpose of the Spark demo from your point of view? Should this be an "Example" as well? Is this documented somewhere? Would it be an option to move it to `/demo` instead of `/docker_build` as it does not seem to actually build an `anovos` Docker container that can run arbitrary `anovos` code?

Yes the purpose of spark demo is written to be an example run as well. It is more to show an End to End run of all functionalities under the different modules in Anovos from data ingestion to data report . yes `/demo` is better. `/docker_build` seems misleading as no container that run arbitrary anovos code is being built.

ionicsolutions:
> > @mw-vishnug
> > Two questions:
> > 
> > 1. Why does the Makefile need specific data? And what is the purpose of the Makefile? Is it actually building the project or merely triggering the tests?
> 
> Agree that the Makefile should be generic and not for any specific data. Anovos was written to run as a pyspark application and therefore the current Makefile on ` build` in `make clean build` builds the "anovos.zip" package to be included while doing spark-submit. Other instructions set is to copy the `src/main/main.py` and the `config/configs.yaml` to the `dist/` folder, which has the configuration to be modified by the user for the user dataset. Additionally, currently it also copies the `data/income_dataset` as a sort of "default dataset" to do the standalone run in local environment. This part I think can be removed and the default `config/configs.yaml` file can be modified to point to the data from the new path `../example/data` so Makefile does not have to be tied to any dataset.

Is this documented somewhere? Based on my current understanding, it might be an option to remove the Makefile from the top-level directory and provide samples and instructions for users to write their own?

> > 1. What exactly is the purpose of the Spark demo from your point of view? Should this be an "Example" as well? Is this documented somewhere? Would it be an option to move it to `/demo` instead of `/docker_build` as it does not seem to actually build an `anovos` Docker container that can run arbitrary `anovos` code?
> 
> Yes the purpose of spark demo is written to be an example run as well. It is more to show an End to End run of all functionalities under the different modules in Anovos from data ingestion to data report . yes `/demo` is better. `/docker_build` seems misleading as no container that run arbitrary anovos code is being built.

OK. Is this documented somewhere? We could have a README in the /demo folder. If we need to access the data during the docker build step, we'll need to include it in the build context or explicitly download it (rather than storing it within the Git repo). Since the Spark demo only runs within Docker (in contrast to the notebooks and guides that users can spin up locally), it might be an option to store the required data in an S3 bucket like you do with the Anovos logo for the reports and then fetch it during the docker build step. That's a pattern that I see and use frequently.

mw-vishnug (approved):
changes are good to be merged

Issue #9, opened 2021-11-11
Outlier detector issues warning that skewed columns are dropped even if there are no skewed columns


ionicsolutions:
The warning message in [quality_checker.py#L483](https://github.com/anovos/anovos/blob/f5123f23905cb28b05fae38557e078141d0a5def/src/main/anovos/data_analyzer/quality_checker.py#L483) should only be issued if `skewed_cols` is not empty.

PR #16, opened 2021-11-15
Only show warning if there are any skewed columns

ionicsolutions:
This warning message shows up in the new Getting Started guide even though there are no skewed columns to name, resulting in a potentially confusing warning message


nisha20verma:
The fix has already been merged into the main as the part of issue #9

ionicsolutions:
@nisha20verma The error still occurs in the Getting Started guide, the check whether there are any skewed columns is not present in `main` as you can see from the diff of this branch to the main branch.

PR #23, opened 2021-11-25
changes for easy api integration changes

mw-vishnug:
1. duplicate_detection:
   - Added number of duplicate rows and percentage of duplicate rows for to print
   
2. IDness_detection:
   - If treatment is true,  the column name 'flagged' will be renamed as 'treated'

3. biasedness_detection:
   - also added 'mode_rows' to select as part of odf_print (Note: Also changed to add 'mode_rows' correspondingly in measures_of_centralTendency.)
   - If treatment is true,  the column name 'flagged' will be renamed as 'treated'

4. Corresponding changes in report_generation

5. outlier_detection changed new line formatting as there was a stray print "skewed_cols))"

### DIFF
--- a/src/main/anovos/data_analyzer/quality_checker.py
+++ b/src/main/anovos/data_analyzer/quality_checker.py
@@ -27,7 +27,8 @@ def duplicate_detection(spark, idf, list_of_cols='all', drop_cols=[], treatment=
     :param treatment: Boolean argument – True or False. If True, duplicate rows are removed from the input dataframe.
     :return: (Output Dataframe, Metric Dataframe)
               Output Dataframe is de-duplicated dataframe if treated, else original input dataframe.
-              Metric Dataframe is of schema [metric, value] and contains metrics - number of rows & number of unique rows.
+              Metric Dataframe is of schema [metric, value] and contains metrics - number of rows, number of unique rows, 
+              number of duplicate rows and percentage of duplicate rows in total.
     """
     if list_of_cols == 'all':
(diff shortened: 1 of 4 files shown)


src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -51,12 +51,16 @@ def duplicate_detection(spark, idf, list_of_cols='all', drop_cols=[], treatment=
...
+    odf_print = spark.createDataFrame([["rows_count", float(idf.count())], \
+					["unique_rows_count", float(odf_tmp.count())], \
+					["duplicate_rows", float(idf.count() - odf_tmp.count())], \
+					["pct_duplicate_rows", round((idf.count() - odf_tmp.count())/idf.count(), 4)]], \
+					schema=['metric', 'value'])
     if print_impact:
         print("No. of Rows: " + str(idf.count()))
         print("No. of UNIQUE Rows: " + str(odf_tmp.count()))
+        print("No. of Duplicate Rows: " + str(odf_tmp.count() - idf.count()))
+        print("Percentage of Duplicate Rows: " + str((odf_tmp.count() - idf.count())/idf.count()))
```
nisha20verma, line 63: - Round off to be consistent with output dataframe. 
- Docstring to be updated wrt metric dataframe
- duplicate_pct instead of pct_duplicate_rows? similar to mode_rows & mode_pct?

nisha20verma: Basic report generation needs to be updated too.

PR #26, opened 2021-12-06
First workflow to execute the demo run

ionicsolutions:
Two basic workflows to be executed on each push to the repository:
- Set up the current default environment and run the Spark demo
- Set up the current default environment and run all unit tests

### DIFF
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -27,6 +27,10 @@ jobs:
         run: |
           make clean build
 
+      - name: Install Python requirements
+        run: |
+          pip install -r requirements.txt
+
       - name: Run the demo
         run: |
           cd dist/
--- a/.github/workflows/demo.yml
+++ b/.github/workflows/demo.yml
@@ -30,9 +30,4 @@ jobs:
       - name: Run the demo
         run: |
           cd dist/
-          ./spark-submit.sh > run.txt
-          tail -n 30 -f run.txt
-          
-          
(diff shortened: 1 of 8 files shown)


ionicsolutions:
The demo runs for a really long time and will eventually keep all available runners busy, preventing additional workflows from being executed until the demos are finished.

Moving forward, it would be nice if the demo would run for just a couple of minutes. I believe this would also be beneficial for users who are trying out Anovos on their local machines.

The unit tests are failing due to issues with the test data. However, the workflow itself is fully functional and also reports coverage (currently only for the files that are touched by the tests, so the 39% does not reflect the actual state).

mw-vishnug:
I changed the current one as full demo run (to run only on pull-request)
and changed the demo as for basic-report (which now runs only for 6mins)
All the checks have passed.
@ionicsolutions  do let know if these flows are good to proceed with , thanks

ionicsolutions:
@mw-vishnug Thanks, looks good to me.

ionicsolutions:
Can this be merged?

Vishnu-Gowthem (approved):
I am good with the unit test, full-demo and demo changes for 3.7 till https://github.com/anovos/anovos/pull/26/commits/5b17d8c2eaa825d4ec58c7925fda6171c6c3f702

ionicsolutions:
We could split this up into two PRs and keep the matrix build as WIP for now?

PR #33, opened 2021-12-13
Refactor quality_checker.duplicate_detection

ionicsolutions:


src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -30,40 +44,38 @@ def duplicate_detection(spark, idf, list_of_cols='all', drop_cols=[], treatment=
...
 
-    if any(x not in idf.columns for x in list_of_cols) | (len(list_of_cols) == 0):
-        raise TypeError('Invalid input for Column(s)')
-    if str(treatment).lower() == 'true':
-        treatment = True
-    elif str(treatment).lower() == 'false':
-        treatment = False
-    else:
-        raise TypeError('Non-Boolean input for treatment')
+    drop_columns = drop_columns or []
```
ionicsolutions, line 53: As a general rule: Never use mutable types as default arguments. The default value is instantiated only once at import time, thus there will be data leakage across function calls.

The pattern `arg = arg or list` as shown here is common practice.

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -30,40 +44,38 @@ def duplicate_detection(spark, idf, list_of_cols='all', drop_cols=[], treatment=
...
+                                       ["unique_rows_count", float(output_df.count())],
+                                       ["duplicate_rows", float(input_df.count() - output_df.count())],
+                                       ["duplicate_pct", round((input_df.count() - output_df.count()) / input_df.count(), 4)]],
+                                      schema=['metric', 'value'])
     if print_impact:
-        print("No. of Rows: " + str(idf.count()))
-        print("No. of UNIQUE Rows: " + str(odf_tmp.count()))
-        print("No. of Duplicate Rows: " + str(idf.count() - odf_tmp.count()))
-        print("Percentage of Duplicate Rows: " + str(round((idf.count() - odf_tmp.count())/idf.count(),4)))
+        print(report_df)
```
ionicsolutions, line 73: There is no reason to compute the data twice. Otherweise, if we're not careful during future changes, the printed report will show different values from the returned report DataFrame.

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -30,40 +44,38 @@ def duplicate_detection(spark, idf, list_of_cols='all', drop_cols=[], treatment=
...
+    drop_columns = drop_columns or []
+    drop_columns = _parse_columns(drop_columns)
+    if drop_columns:
+        columns = [col for col in columns if col not in drop_columns]
+
+    if not columns:
+        raise ValueError("No columns to work on.")
 
-    odf_tmp = idf.drop_duplicates(subset=list_of_cols)
-    odf = odf_tmp if treatment else idf
```
ionicsolutions, line 53: Instead of introducing a "temporary" variable, we can decide what to return at the end of the function

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -9,19 +10,32 @@
...
+    if isinstance(list_of_cols, str):
+        list_of_cols = [x.strip() for x in list_of_cols.split('|')]
+    return list(set(list_of_cols))
+
+
+def duplicate_detection(spark: SparkSession,
+                        input_df: DataFrame,
+                        columns: Union[List[str], str] = 'all',
+                        drop_columns: Optional[Union[List[str], str]] = None,
+                        *,
```
ionicsolutions, line 26: Make the two flags explicit by forcing users to specify them as keyword arguments.

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -9,19 +10,32 @@
 from anovos.shared.utils import attributeType_segregation, transpose_dataframe, get_dtype
 from pyspark.sql import functions as F
 from pyspark.sql import types as T
+from pyspark.sql import SparkSession, DataFrame
 
 
-def duplicate_detection(spark, idf, list_of_cols='all', drop_cols=[], treatment=False, print_impact=False):
+def _parse_columns(list_of_cols: Union[List[str], str]):
```
ziedbouf, line 16: why are we using typing in the context

ziedbouf: the rest of project doesn't include any typing

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -9,19 +10,32 @@
...
+    if isinstance(list_of_cols, str):
+        list_of_cols = [x.strip() for x in list_of_cols.split('|')]
+    return list(set(list_of_cols))
```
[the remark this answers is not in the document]
ziedbouf: i agree, however i don't see the necessity for such a function. Is it possible to explain why we need this?

src/main/anovos/data_analyzer/quality_checker.py
ziedbouf, line 16: what do you think of adding delimiter or seperator part pf the function with default value |,
`def _parse_columns(list_of_cols: Union[List[str], str], delimiter="|"):`

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -236,7 +248,7 @@ def nullColumns_detection(spark, idf, list_of_cols='missing', drop_cols=[], trea
         treatment_threshold = float(treatment_threshold)
     else:
         if treatment_method == 'column_removal':
```
ziedbouf, line 250: cannot see the `treatment_method` part of the the function signature

src/main/anovos/data_analyzer/quality_checker.py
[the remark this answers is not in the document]
ionicsolutions: @VishnuGowthem I believe that you know the answer here :-)

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -1,6 +1,7 @@
 # coding=utf-8
 import re
 import warnings
+from typing import List, Union, Optional, Tuple
 
 from anovos.data_analyzer.stats_generator import uniqueCount_computation, missingCount_computation, mode_computation, \
```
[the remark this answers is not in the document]
ionicsolutions: Agree. That's what we enforce with the new `black` code formatter :-)

src/main/anovos/data_analyzer/quality_checker.py
ionicsolutions: Is there ever a case where the delimiter is different? @VishnuGowthem

ziedbouf: no totally agree, but will be better first just to define the types without enforcing python typing, and we can rollback second version with generalized typing:

```def duplicate_detection(spark: Sparksession, idf: Dataframe, list_of_cols: [] ='all', drop_cols: list =[], treatment: bool =False, print_impact: bool =False) -> Dataframe:
```

No idea if what i mentioned above works, need to check it

PR #37, opened 2021-12-19
WIP: Compatibility with newer versions of Spark

ionicsolutions:


ionicsolutions:
Moving beyond Python 3.7 is not possible due to the dependency on `pyarrow==3.0.0`

There seems to be a general problem with how the Spark session is set up during tests, causing them to be flaky. This is also true for the current stable configuration with Spark 2.4.8 and should be addressed when revising the session management.

kajanansangar:
@ionicsolutions we have made a branch that is runnable on Spark3 & python 3.8 (refer spark_upgrade_3_link)

ionicsolutions:
Superseded by #40

PR #40, opened 2022-01-03
WIP: Upgrade to Spark 3

ionicsolutions:


bin/spark-submit.sh
```diff
@@ -15,14 +15,14 @@ spark-submit \
...
 --conf spark.executor.extraJavaOptions="-Dlog4j.configuration=file://$SPARK_HOME/conf/log4j.properties" \
 --conf spark.driver.extraJavaOptions="-Dlog4j.configuration=file://$SPARK_HOME/conf/log4j.properties" \
---conf spark.kryo.referenceTracking=false \
+--conf spark.kryo.referenceTracking=true \
 --conf spark.network.timeout=18000s \
 --conf spark.executor.heartbeatInterval=12000s \
 --conf spark.dynamicAllocation.executorIdleTimeout=12000s \
 --conf spark.port.maxRetries=200 \
---packages org.apache.spark:spark-avro_2.11:2.4.0 \
+--packages org.apache.spark:spark-avro_2.12:3.1.2 \
```
ionicsolutions, line 23: @dattranm This will not work for Spark 2.4.x

Do you have a suggestion how to adapt spark_submit.sh depending on the Spark version used?

ionicsolutions:
The unit tests run in suitable configurations now. Failure to run on Spark 3 will **not** cause the workflow to fail.

The demo runs do not work yet, which appears to be due to some subtle issue with the Spark configuration. We need to find a sensible way to maintain the two different variants of `bin/spark-submit.sh`. Either by auto-detecting the Spark version within the shell script (preferred) or by providing two separate files.

ionicsolutions:
@VishnuGowthem Could you have a look at the questions surrounding the `spark-submit.sh` script above, please? We should find a way to either detect the Spark version within the script (as we do in [`shared.spark`](https://github.com/anovos/anovos/blob/d84e45d60ea8ecc679ad667a304e191aa09e457c/src/main/anovos/shared/spark.py#L13)) or provide two separate versions

ionicsolutions:
The demo run for Spark 3.2.0 fails with

```
java.lang.NoSuchMethodError: 'org.apache.spark.internal.config.ConfigEntry org.apache.spark.sql.internal.SQLConf$.LEGACY_AVRO_REBASE_MODE_IN_READ()'
```

This points to an Avro incompatibility, see [this SO post](https://stackoverflow.com/questions/61511362/java-lang-nosuchmethoderror-when-reading-an-avro-file-using-pyspark).

Which version of Spark did you test with internally/locally? The package `org.apache.spark#spark-avro_2.12` is currently loaded for Spark 3.1.2, there [is a version for 3.2.0](https://repo1.maven.org/maven2/org/apache/spark/spark-avro_2.12/) as well. It might be a good idea to always load the version of `spark-avro` specifically for the Spark version used?

ionicsolutions:
The demo run for Spark 2.4.8 fails with

```
dp.Text(" Categorical Attributes Name : **" + str(catcols_name) + "**"), rows=6), rows=8)
```

As far as I can tell, `dp.Text` never had a `rows` parameter ([this commit seems to have introduced the `Text` element](https://github.com/datapane/datapane/commit/6fade3bebf2c470befa9cd799b6a286f048fa80e#diff-a8b67ceea20aa8d04b4434fbda13019e2dc513356cd6002ccec34e0ffd64b962R341)) but `dp.Group` [certainly does](https://github.com/datapane/datapane/blob/50d9f6a5634140e17cfd07244c4371cbd8001d4a/src/datapane/client/api/report/blocks.py#L278) and I believe that's the class that's initialized here:

```python
    l1 = dp.Group(
        dp.Text("# "), \
        dp.Text("*This section summarizes the dataset with key statistical metrics.*"), \
        dp.Text("# "), \
        dp.Text("# "), \
        dp.Text("### Global Summary"), \
        dp.Group(dp.Text(" Total Number of Records: **" + str(f"{rows_count:,d}") + "**"),
                 dp.Text(" Total Number of Attributes: **" + str(columns_count) + "**"),
                 dp.Text(" Number of Numerical Attributes : **" + str(numcols_count) + "**"),
                 dp.Text(" Numerical Attributes Name : **" + str(numcols_name) + "**"),
                 dp.Text(" Number of Categorical Attributes : **" + str(catcols_count) + "**"),
                 dp.Text(" Categorical Attributes Name : **" + str(catcols_name) + "**"), rows=6), rows=8)
```

VishnuGowthem:
@ionicsolutions thanks much for the pointers. 

changed the spark-submit script and spark.py to take specific versions for spark for Avro. and datapane current changes are good if specifically == 0.12.0 are taken instead of >= 0.12.0.. currently have restricted it.. 
will check if we can make code changes to update to latest version..

PR #45, opened 2022-01-18
Check for compatibility with all supported versions of Spark 3.x

ionicsolutions:


ionicsolutions:
I added a demo run for Spark 3.1.2 prior to updating the docs and discovered in the process that since last night all our unit tests break because findspark was updated to 2.0.0 and this version breaks our unit tests. I fixed the version in the requirements.txt.

kajanansangar:
I am getting the following error while I am running the "make clean build test" in my local for this branch

Refer to the paste bin logs [here](https://pastebin.com/LUmszaR9) 

Issue persists in the main branch too. Refer to the paste bin logs [here](https://pastebin.com/CRGA3vV3) 

@ionicsolutions @Vishnu-Gowthem did we check this before we merge PR #40

VishnuGowthem (approved):
I am good with these changes.

ionicsolutions:
@kajanansangar This branch does not change anything about the library itself, so I don't believe any of the changes here are the cause of the issue you encounter. From what I can see, you are on Spark 2.4.7, which we don't check for in any of the workflows.

The unit tests and the demo run fine on 2.4.8 (with Python 3.7) as well as 3.1.2 and 3.2.1 (with Python 3.9): https://github.com/anovos/anovos/pull/45/checks

You might encounter the same issue with the latest `findspark` update that we had here. How do you install your dependencies and from which source do you install them?

kajanansangar (approved):
@ionicsolutions I have checked the same with @VishnuGowthem  and found out that issue is due to pip3 is having version 2.0.0 of "findspark". I have installed the lower version of "findspark" only on my "pip". I am able to run E2E now and I am merging it now. Thanks guys.

PR #54, opened 2022-01-25
Azure databricks testing

ionicsolutions:


src/main/anovos/data_report/basic_report_generation.py
```diff
@@ -48,7 +48,7 @@ def anovos_basic_report(spark, idf, id_col='', label_col='', event_label='', out
     :param label_col: Label/Target column
     :param event_label: Value of (positive) event (i.e label 1)
     :param output_path: File Path for saving metrics and basic report
-    :param local_or_emr: "local" (default), "emr"
+    :param global_run_type: "local" (default), "emr"
                          "emr" if the files are read from or written in AWS s3
```
ionicsolutions, line 52: @ranjanravish Can you please add a short documentation for the new "databricks" option? Thanks!

src/main/anovos/data_report/basic_report_generation.py
```diff
@@ -62,9 +62,11 @@ def anovos_basic_report(spark, idf, id_col='', label_col='', event_label='', out
     AA_funcs = [correlation_matrix, variable_clustering]
     AT_funcs = [IV_calculation, IG_calculation]
     all_funcs = SG_funcs + QC_rows_funcs + QC_cols_funcs + AA_funcs + AT_funcs
-
-    if local_or_emr == "local":
+    
+    if global_run_type == "local":
         local_path = output_path
+    elif global_run_type == "databricks":
+        local_path = "/dbfs/FileStore/tables/report_stats"
```
ionicsolutions, line 69: @ranjanravish Is this `local_path` something a user might want to change?

src/main/anovos/data_report/basic_report_generation.py
```diff
@@ -62,9 +62,11 @@ def anovos_basic_report(spark, idf, id_col='', label_col='', event_label='', out
...
     else:
```
ionicsolutions, line 70:
```suggestion
    elif global_run_type == "emr":
        ...
    else:
        raise ValueError("Invalid global_run_type")
```

We should only allow supported run types and raise a ValueError to catch typos etc. and explicitly tell the user what's going wrong. Otherwise, they might accidentally type "databircks" and wonder why their workflow doesn't run as expected.

src/main/anovos/data_report/basic_report_generation.py
```diff
@@ -62,9 +62,11 @@ def anovos_basic_report(spark, idf, id_col='', label_col='', event_label='', out
...
         local_path = "report_stats"
     Path(local_path).mkdir(parents=True, exist_ok=True)
```
ionicsolutions, line 72: Does this imply we silently overwrite old output files?

src/main/anovos/data_report/report_preprocessing.py
```diff
@@ -32,9 +32,11 @@ def save_stats(spark, idf, master_path, function_name, reread=False, run_type="l
     :run_type: local or emr based on the mode of execution. Default value is kept as local
     :reread: option to reread. Default value is kept as False
     """
-
+    
     if run_type == "local":
         local_path = master_path
+    elif run_type == "databricks":
+        local_path = "/dbfs/FileStore/tables/report_stats"
     else:
```
ionicsolutions, line 40: @ranjanravish As above, I'd explicitly check for all types using `if`/`elif` and raise a `ValueError` if the user gives a different one. Otherwise, there'll be lots of silent failures that are hard to track down.

src/main/anovos/data_report/report_preprocessing.py
```diff
@@ -300,9 +302,11 @@ def charts_to_objects(spark, idf, list_of_cols='all', drop_cols=[], label_col=No
     cutoffs_path2 = source_path + "/drift_statistics/attribute_binning"
 
     idf_encoded.persist(pyspark.StorageLevel.MEMORY_AND_DISK)
-
+    
     if run_type == "local":
         local_path = master_path
+    elif run_type == "databricks":
+        local_path = "/dbfs/FileStore/tables/report_stats"
     else:
```
ionicsolutions, line 310: @ranjanravish Same here, don't silently to an `else` if a user misspells "databricks"

src/main/main.py
```diff
@@ -248,26 +248,26 @@ def main(all_configs, local_or_emr):
...
 
 if __name__ == '__main__':
     config_path = sys.argv[1]
-    local_or_emr = sys.argv[2]
+    global_run_type = sys.argv[2]
 
-    if local_or_emr == 'local':
+    if global_run_type == 'local' or 'databricks':
         config_file = open(config_path, 'r')
     else:
```
ionicsolutions, line 267: @ranjanravish Same reasoning applies here as well :-)

ionicsolutions:
I cherry-picked the merge commit of #45 on this branch to ensure that all tests pass and added a review with some comments regarding potential causes of silent failures due to typos etc. in the user's input.

ionicsolutions:
@ranjanravish The demo runs (and the code quality check) fail due to a syntax error in main.py. Could you please have a look? Thanks!

src/main/anovos/data_report/basic_report_generation.py
ranjanravish: Added short documentation for the new "databricks" option in docstring

src/main/anovos/data_report/basic_report_generation.py
ranjanravish: yeah, i removed this static path and added function for assigning this local path

src/main/anovos/data_report/basic_report_generation.py
ranjanravish: Added condition that will raise a ValueError to catch typos

src/main/anovos/data_report/report_preprocessing.py
ranjanravish: Here also, Added condition that will raise a ValueError to catch typos

src/main/anovos/data_report/report_preprocessing.py
ranjanravish: Added condition that will raise a ValueError to catch typos

PR #57, opened 2022-01-28
PyPI release workflow

ionicsolutions:


ionicsolutions:
@VishnuGowthem Can you check why authentication with test.pypi.org fails? https://test.pypi.org/help/#invalid-auth

ziedbouf:
@ionicsolutions  i am working on setting up poetry for dependency management, this might simplify the CI process and publishing the work to pypi, will try to finalize it and publish the pull request soon. you can look into [github ci best practice with poetry](https://github.com/python-poetry/poetry/issues/366#issuecomment-691412462)

ziedbouf:
@ionicsolutions this pull request #64 under progress but i think we might take advantage of poetry to manage dependency and publish to pypi

ionicsolutions:
@ziedbouf Thanks for the suggestion. I left a longer comment on #64 detailing why I'm very hesitant to switch to `poetry` at this point.

VishnuGowthem:
> @VishnuGowthem There still seems to be an issue with the authentication:
> 
> ```
> Warning:  It looks like you are trying to use an API token to authenticate in the package index and your token value does not start with "pypi-" as it typically should. This may cause an authentication error. Please verify that you have copied your token properly if such an error occurs.
> ```
> 
> https://github.com/anovos/anovos/runs/5066207262?check_suite_focus=true#step:5:11

I had missed to mention the environment for the publish block. now there is a different error 

Error during upload. Retry with the --verbose option for more details.
[24](https://github.com/anovos/anovos/runs/5091046873?check_suite_focus=true#step:5:24)
HTTPError: 400 Bad Request from https://test.pypi.org/legacy/
[25](https://github.com/anovos/anovos/runs/5091046873?check_suite_focus=true#step:5:25)
File already exists. See https://test.pypi.org/help/#file-name-reuse for more information.

Think it is because the version number is the same and it conflicts with the existing file

VishnuGowthem:
The .whl file has got updated on test Pypi but the tgz upload had errored as we had the file name with the same version
 
image

ionicsolutions:
@VishnuGowthem OK, interesting, looks like we can only test-publish once. I'll have a closer look later today.

ionicsolutions:
@VishnuGowthem Aside from the versioning problem, we run into [A file with the exact same content exists](https://test.pypi.org/help/#file-name-reuse) which is fine I guess.

ionicsolutions:
As discussed in today's meeting:

PR #60, opened 2022-02-01
Add code quality and formatting checks

ionicsolutions:


ionicsolutions:
@kajanansangar I added basic linting with `flake` (which already caught an undefined variable) as well as more advanced checks with `black`.

Moving forward, once we have more comprehensive testing in place, we can structure the workflows as `linting -> unit test -> demo runs` but for now I believe it's worthwhile to just keep them running in parallel because they're all checking different things

kajanansangar (approved):
the formatting looks good to me. @ionicsolutions what code formatting style it follows?

ionicsolutions:
> the formatting looks good to me. @ionicsolutions what code formatting style it follows?

@kajanansangar It uses the defaults of `black`, see https://black.readthedocs.io/en/stable/the_black_code_style/index.html

Personally, I think it's best to just go with the defaults, otherwise there will be lots of discussions and opinions that don't really help ;-)

PR #61, opened 2022-02-01
Remove obsolete API specification from main repo

ionicsolutions:
The API docs are now auto-generated in https://github.com/anovos/anovos-docs and shipped with the main documentation to https://docs.anovos.ai


kajanansangar (approved):
good to be removed

PR #64, opened 2022-02-06
move to poetry for dependency management

ziedbouf:
- [x] move setup.py and setup.cfg to pyproject.toml
- [x] fix docker file demo to install through poetry
- [ ] fix examples and build using poetry
- [ ] add release workflow using poetry
- [ ] test the workflow


ionicsolutions:
Hi @ziedbouf, thanks for your suggestion. It's not my call to make but I don't think it's the right point in time and the right decision currently to switch to `poetry`.

On the one hand, we just started moving towards a more robust and maintainable dependency management. It will take quite some time go get this right and I'm very hesitant to introduce yet another new tool. Further, I don't think the current workflow for using Anovos is compatible with `poetry`.

On the other hand, the focus until the 1.0 release due this summer is on turning Anovos into a library that's suitable to be used outside of Mobilewalla. We're slowly establishing foundations like uniform code styles, security scanning and, perhaps most importantly, automated and more comprehensive testing. There are no issues with using the conventional `pip` and `setup.py/cfg` structure and we rely on a `requirements.txt` in many situations. (In fact, we're just now in the process of consolidating all requirements into this one text file.)

However, as I said, I'm not the one making these decisions, so let's hear what @kajanansangar thinks :)

ziedbouf:
Thanks @ionicsolutions  it's fully understandabale, will keep playing locally with the setup for poetry. In the meanwhile, i will try to pick up something else for contribution. @kajanansangar  please feel free to discard the current pull request if it doesn't fit anovos roadmap. Also, is there a feature pool that we can pick up from?

ionicsolutions:
@ziedbouf I think the biggest contribution right now would be to help with building up the unit/integration test suite. The test coverage right now is really low, which is preventing larger refactorings and is causing issues whenever dependencies are to be updated, because the tests and demos only execute parts of the library.

If you have ideas how to (better) set up comprehensive testing, making full use of `pytest` etc., feel free to provide a demonstration just as you did with this PR.

If you're the kind of person that likes to debug and fiddle with dependencies and workflows, you could have a look at #50 first. For some reason, `findspark` 2.0.0 does not work for the current Spark setup in the testing workflows. It could be a great starting point for an overhaul of `conftest.py`

ionicsolutions:
In addition, there are also about 500 warnings issued by `flake`, many of them pointing to potentially undefined variables and imports: https://github.com/anovos/anovos/runs/5087750749?check_suite_focus=true#step:4:50

I know it's not the most attractive/impressive work to work on that but I believe it's an important step to get Anovos in shape for future developments.

ziedbouf:
interesting @ionicsolutions, will look into the testing and will try to spin up few test cases and see how it goes

kajanansangar:
@ziedbouf thanks for your intention and understanding. I feel & agree with @ionicsolutions sentiments. Keep up your good work @ziedbouf

PR #67, opened 2022-02-08
fix linter detected warning related to format, unused variables and i…

ziedbouf:
…gnore non python files

- [x] add execuled filees/folders to flake8 cmd
- [x] use absolute imports statements
- [x] rename few variables `local_or_emr -> local_or_cluster`, `l -> target_association_rep`, `i -> _idx`, etc...
- [x] remove few unused variables
- [x] add print_impact to few docstrings
- [x] few changes under `src/test`
- [x] add `dev_requirements.txt` for development dependency 
- [x] add loguru for logging 
- [x] hanging yaml import, i think we are mainly using PyYaml but need to cross check with team member
- [ ] few indentation still hanging out, the best to way to resolve it will be through pre-commit hooks

This shall drops warning from 500 to around 40


ziedbouf:
@ionicsolutions check this pull request, it's pretty large but straight forward. Only fixing imports, format and renaming. Also i added `dev_requirements.txt` files

ziedbouf:
@ionicsolutions  hit an issue related to black exit with code 1, but i need an approval in order for me to run the job and try to debug the issue

ionicsolutions:
@ziedbouf I don't think I have the necessary permissions, so @kajanansangar or @VishnuGowthem will need to have a look :)

ziedbouf:
@ionicsolutions  any idea why black formatting fails, cannot see error details and the action execution https://github.com/ziedbouf/anovos/runs/5112008390?check_suite_focus=true

ionicsolutions:
> @ionicsolutions any idea why black formatting fails, cannot see error details and the action execution https://github.com/ziedbouf/anovos/runs/5112008390?check_suite_focus=true

There's a newline missing here (the plus sign indicates an added newline): https://github.com/ziedbouf/anovos/runs/5112008390?check_suite_focus=true#step:5:53

But it looks like you removed the dependency on `findspark` anyway, so this is obsolete?

.github/workflows/lint.yml
```diff
@@ -21,12 +21,13 @@ jobs:
...
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics --exclude=.git,__pycache__,data,demo,jars,.idea,venv,.tox,.parquet,.json.gz,.csv
 
       - name: Lint with Black
         uses: psf/black@stable
         with:
           options: "--check --verbose --diff"
           src: "./src/main"
+          exclude: ^./src/main/anovos/shared/spark.py
```
ionicsolutions, line 33: I don't think that's necessary. Is this still a problem after you run `black src/`?

.github/workflows/lint.yml
```diff
@@ -21,12 +21,13 @@ jobs:
           pip install --upgrade pip
           pip install flake8
           # stop the build if there are Python syntax errors or undefined names
-          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics
+          flake8 . --count --select=E9,F63,F7,F82 --show-source --statistics --exclude=.git,__pycache__,data,demo,jars,.idea,venv,.tox,.parquet,.json.gz,.csv
           # exit-zero treats all errors as warnings. The GitHub editor is 127 chars wide
-          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics
+          flake8 . --count --exit-zero --max-complexity=10 --max-line-length=127 --statistics --exclude=.git,__pycache__,data,demo,jars,.idea,venv,.tox,.parquet,.json.gz,.csv
```
ionicsolutions, line 26: Is this necessary when we have a `.flake8`? Or a flake8 section in the `pyproject.toml`?

requirements.txt
```diff
@@ -11,9 +11,13 @@ s3path
 pybind11
 popmon
 varclushi
-pytest
 cython
 soupsieve
 fsspec
 s3fs
 findspark<2.0.0
+pyspark>=2.4.3,<=2.4.8
```
ionicsolutions, line 19: This breaks compatibility with Spark 3.x, no?

ionicsolutions (changes requested):

requirements.txt
ziedbouf: yes but i got this is requirement when i scanned the project using poetry, i think it has conflict with some libray we call for

.github/workflows/lint.yml
ziedbouf: both shall works, honestly didn't check pyproject.toml. It can goes also under setup.cfg

.github/workflows/lint.yml
ziedbouf: locally works fine, i suspected it's related to imports order with  `findspark.init()` before pyspark imports. no idea if order matter, as i need to check it first

requirements.txt
ionicsolutions: We just introduced compatibility with Spark 3.x, so that's not an option :-)

Which package has the issues? Would be good to know! The demos and unit tests run fine under several Spark versions, but of course that's an imperfect test.

.github/workflows/lint.yml
ionicsolutions: At this point, I'm fine with putting it in one of these places, but we should not exclude folders in the workflow as well. They should be defined in one place only :-)

ionicsolutions:
@ziedbouf Could you please rebase your branch to the current `main` and force-push? This would resolve the conflicts in a clean way :-)

.github/workflows/lint.yml
ionicsolutions: Yes, please check this. We should not exclude arbitrary files just because some checks fail. If we exclude files, it should be for well-understood and documented reasons.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -1,13 +1,11 @@
 ---
-name: Bug report
-about: Create a report to help us improve
-title: ''
+name: Bug report about: Create a report to help us improve title: ''
 labels: ''
 assignees: ''
 
 ---
```
ionicsolutions, line 6: That doesn't seem right?

ziedbouf: the main branch has diverged with huge difference but what's have been on my end and what's have been committed, still looking to the code and trying to resolve and fix issues.

ionicsolutions: Hmm, there are not many changes to the files in `src` I hope?

I'd like to try to get this PR merged prior to the feature branches that are currently open. It would help the team to see what you changed to avoid the various issues raised by `flake` and adapt their code accordingly. (Rather than you having to fix a new round of issues.) Super helpful what you're doing!

ziedbouf: yes i totally agree as so many things are format related, still many possible enhancment that we can get into the code. Just working on my setup :D and pycharm hits me with few wired imports :'(

requirements.txt
ziedbouf: checked again poetry and the following shall be supported:
image

ziedbouf: @ionicsolutions  will update the requirements for pyspark to cover `3.2.1`

ionicsolutions: My current understanding is that `pyspark` is installed through Spark and not via `pip`. As it needs to match whatever Spark version a user has installed, I don't think we should add it in `requirements.txt`.

ionicsolutions: Have a look at https://github.com/anovos/anovos/blob/main/.github/workflows/demo.yml for example

ionicsolutions: That's a good idea. We're working on the dependencies anyway, some have been fixed to arbitrary versions for unknown reasons. The only thing we know we have issues with are datapane #48 (there was a breaking change between 0.12.x and 0.13.x) and findspark #50 (which is just a test requirement and I don't know why 2.0.0 doesn't work for us).

ziedbouf:
@ionicsolutions forced black to ignore top level imports for findspark `#  noqa: E402`, and i am hitting another one. My concern that those rules are not forced on my local setup and i am surprised why they are raised within the the github action;
https://github.com/ziedbouf/anovos/runs/5126377007?check_suite_focus=true

ionicsolutions:
> @ionicsolutions forced black to ignore top level imports for findspark `# noqa: E402`, and i am hitting another one. My concern that those rules are not forced on my local setup and i am surprised why they are raised within the the github action; https://github.com/ziedbouf/anovos/runs/5126377007?check_suite_focus=true

I think you're mixing up things here. Black does not care about imports, it just checks the code formatting. The error you linked has to do with spaces around the `**` operator, not with any imports. Black will format code the same everywhere when you run it.

ionicsolutions:
Also, it's not our goal to necessarily silence all `flake` warnings. If there's a good reason to deviate from the standard, we can do that.

ziedbouf:
interesting @ionicsolutions and i was really confused as everything worked fine under my setup and didn't catch the issue, thanks for the help. Right now the pull request is in good shape. Few trailing white-spaces and lines that exceed the line limits.

ionicsolutions (approved):

kajanansangar:
@ziedbouf good job on your first merged PR. Congratulations & appreciations. Keep up !!

PR #82, opened 2022-02-15
Transformers: Final round of changes

nisha20verma:


dattranm (commented):
Most of the changes are related to function and code cleanup, including name changes and/or argument/variable changes. All the logic is still working properly, and it is good to merge now

dattranm:
Only minor change regarding the format of the function and an additional method since last review. Everything here is good to merge.

PR #71, opened 2022-02-10
Refactor drift detection

ziedbouf:
refactoring drift detector and try to enhance argument validation and simplify the code if possible:
- [x] move distance to seperate file
- [x] move argument validation to seperate folder
- [x] check decorator for potential future use
- [x] fix example notebooks
- [x] fix docker build demo if needed

PR #87, opened 2022-02-18
Fix column validator

ionicsolutions:
Fix issues introduced with #71 

- [x] Use `kwargs` in `validate()` (`args` is a tuple, can only be accessed by index)
- [x] Make all relevant arguments of `statistics` keyword-only
- [x] Do not allow mutable default value
- [x] Format code with `black`  
- [x] Workaround for `idf_target` being passed as either `kwarg` or `arg`
- [x] Do not run GitHub Actions jobs that depend on tags when no tag is given 
- [x] Adapt `setup.py`

### DIFF
--- a/src/main/anovos/data_report/report_generation.py
+++ b/src/main/anovos/data_report/report_generation.py
@@ -380,7 +380,7 @@ def drift_stability_ind(
         missing_recs_drift: Missing files from the drift tab
         drift_tab: "drift_statistics"
     missing_recs_stability: Missing files from the stability tab
-        stability_tab:"stability_index_computation, stabilityIndex_metrics"
+        stability_tab:"stability_index, stabilityIndex_metrics"
 
     """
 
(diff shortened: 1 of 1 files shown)


ionicsolutions:
@ziedbouf FYI, your PR #71 caused some issues. Can you please have a look at my bugfixes? Thanks!

src/main/anovos/drift/validations.py
```diff
@@ -18,37 +19,40 @@ def check_list_of_columns(
     def validate(*args, **kwargs):
         logger.debug("check the list of columns")
 
-        if args[columns] == "all" and args[target] is not None:
-            num_cols, cat_cols, other_cols = attributeType_segregation(args[target])
+        cols, drops = [], []
+
+        idf_target = kwargs.get(target, "") or args[target_idx]
```
ionicsolutions, line 24: @ziedbouf This works for now but I don't think that's a viable solution

VishnuGowthem:
I made a fix towards this. kindly help check the same too. Thanks !

> @ziedbouf Can you please check why this error is raised https://github.com/anovos/anovos/runs/5244216338?check_suite_focus=true#step:8:522 ? Thanks!

VishnuGowthem:
All the tests passed, proceeding to merge to main branch

VishnuGowthem (approved):

PR #89, opened 2022-02-21
Time Series Module

sourjyasen07:
Made the changes on the following scripts  as a part of introduction to the timeseries analyzer module - 

- src/main/main.py
- src/main/anovos/data_report/ts_preprocessing.py
- src/main/anovos/data_report/report_generation.py
- config/configs.yaml


ionicsolutions:
@sourjyasen07 Can you please have a look at why the demo runs fail? Thanks!

ionicsolutions:
@VishnuGowthem Let's add some unit tests to isolate the issue and make sure it's fixed. Do you know where exactly the parsing goes wrong?

VishnuGowthem:
> @VishnuGowthem Let's add some unit tests to isolate the issue and make sure it's fixed. Do you know where exactly the parsing goes wrong?

yes, in the latest commit, all the modules passed in local run. 
I missed to make the configs change in configs-basic and configs-full hence the automated workflow has failed.

VishnuGowthem:
The current changes should pass the basic and full-demo runs but the ts_analyzer seems to increase the time taken. 
Attn. @kajanansangar

ionicsolutions:
@VishnuGowthem @sourjyasen07 Do you have an idea why it takes so long? Ideally, the short demo should take between five and ten minutes, as we use it to test all changes right now. Is there a way to reduce the amount of data that is processed?

sinuochen:
I have reviewed the new changes and they look good to me.

kajanansangar (changes requested):
1. Choose the right panel data and keep it as sample data for testing for the time-series module. As of now, there is no sample data provided specifically for this module.
> http://people.stern.nyu.edu/wgreene/Econometrics/PanelDataSets.htm (choose Statewide Capital Productivity Data, lower 48 states, 17 years) as a sample. I would recommend that. Further, the current weather dataset does not have any panel IDs. If we are to use it we need to rename the columns properly (remove _)
2. Create a specific config file only for time-series data
3. Metric dictionary is missing for TS module
4. CRC files are still in the commit

kajanansangar (approved):
good to be merged

kajanansangar:
Finally, one of the heavy modules of Anovos V0.2 is merged. Kudos to @sourjyas @VishnuGowthem @nisha20verma @sinuochen @ionicsolutions

Issue #94, opened 2022-02-23
Importing `feature_exploration` downloads almost 500MB from HuggingFace's model hub


ionicsolutions:
Relevant line: https://github.com/anovos/anovos/blob/main/src/main/anovos/feature_recommender/feature_exploration.py#L6

I suggest to lazy load the model. Otherwise, simply importing the module immediately triggers a download of about 0.5 GB. That's rather surprising for users and not fair to HuggingFace.

kajanansangar:
@dattranm here is the @ionicsolutions's suggestion https://dev.to/thibaultduponchelle/python-lazy-loading-of-an-attribute-aij

Design init_model inside each file and call it explicitly instead of implicit model loading

dattranm:
@ionicsolutions The download only happens once when the users first setting up. After that, it will go into lazy load mode automatically, and no further download is needed once they have the model in local cache.
Though, to address the issue of designing the init_model function and call the model and data explicitly, I have created #97 about some newly created function and properly construct featrec_init. @kajanansangar you can take a look as well.

ionicsolutions:
@dattranm Thanks for addressing the issue!

Sure, there is a cache. But every time a user imports the module for the first time, they are surprised with a >500MB download that they are not informed about and have no way of stopping. Depending on their internet connection and how their cloud provider charges for traffic, that can cause trouble.

Libraries like NLTK and spaCy hence provide separate download functionality and tell users to download the respective asset when they first invoke it. See https://www.nltk.org/data.html and https://spacy.io/usage/models for reference.

The HuggingFace model is also not lazy loaded even if it is cached: It is always loaded on import, which depending on disk speed can take some time and clutters the working memory.

dattranm:
This issue has been resolved with #97 PR, for both adding a download function and lazy load the model.

PR #97, opened 2022-02-23
Feature Recommender minor structure changes

dattranm:
Change to way calling model and input dataframe:
- Create functions to wrap all the data process steps inside
- Create function to call the model implicitly instead of explicitly
- Properly structure the featrec_init to for both feature_exploration and feature_recommendation


src/main/anovos/feature_recommender/featrec_init.py
```diff
@@ -1,6 +1,70 @@
-from anovos.feature_recommender.feature_exploration import *
+from sentence_transformers import SentenceTransformer
 from re import finditer
+import pandas as pd
+import numpy as np
 import copy
+import warnings
+
+
+warnings.filterwarnings("ignore")
```
ionicsolutions, line 9:
```suggestion
```

Please don't hide warnings in library code. If a user doesn't want to see warnings, they can suppress them further up in their own code.

src/main/anovos/feature_recommender/featrec_init.py
```diff
@@ -1,6 +1,70 @@
...
+
+def init_model():
+    """
+
+    Returns
+    -------
+    Semantic model used in Feature Explorer and Recommender
+    """
+    model = SentenceTransformer("all-mpnet-base-v2")
+    return model
```
ionicsolutions, line 20:
```suggestion
class _TransformerModel:
    def __init__(self):
        self._model = None

    @property
    def model(self) -> SentenceTransformer:
        if self._model is None:
            self._model = SentenceTransformer("all-mpnet-base-v2")
        return self._model

model = _TransformerModel()
```

I wouldn't load the model every time, but exactly once. Loading such a huge model even from disk can cause significant delay.

src/main/anovos/feature_recommender/featrec_init.py
```diff
@@ -71,18 +135,53 @@ def recommendation_data_prep(df, name_column, desc_column):
...
+    Returns
+    -------
+    list_train_fer
+        List of prepared data for Feature Recommendation functions
+    df_red_fer
+        DataFrame used in Feature Recommendation functions
+    list_embedding_train_fer
+        List of embedding tensor for Feature Recommendation functions
+    """
+    model_fer = init_model()
```
ionicsolutions, line 163:
```suggestion
```

See above, no need to load the model every time

src/main/anovos/feature_recommender/featrec_init.py
```diff
@@ -71,18 +135,53 @@ def recommendation_data_prep(df, name_column, desc_column):
...
+    list_train_fer, df_rec_fer = recommendation_data_prep(
+        df_groupby_fer, feature_name_column, feature_name_column
     )
-    .reset_index()
-)
-list_train_fer, df_rec_fer = recommendation_data_prep(
-    df_groupby_fer, feature_name_column, feature_name_column
-)
-list_embedding_train_fer = model_fer.encode(list_train_fer, convert_to_tensor=True)
+    list_embedding_train_fer = model_fer.encode(list_train_fer, convert_to_tensor=True)
```
ionicsolutions, line 186:
```suggestion
    list_embedding_train_fer = model.model.encode(list_train_fer, convert_to_tensor=True)
```

ionicsolutions (changes requested):
Thanks for addressing the issue, I've added a suggestion to introduce proper lazy loading

dattranm:
Thanks @ionicsolutions . All of your suggested changes have been committed, including model lazy load and remove the ignore warnings.

src/main/anovos/feature_recommender/feature_recommendation.py
```diff
@@ -222,7 +240,7 @@ def feature_recommendation(
                     df_append = pd.DataFrame(
                         [
                             [
-                                df_user[desc_column].iloc[i],
+                                df_user[name_column].iloc[i],
                                 "Null",
```
kajanansangar, line 244: Null should be replaced with NA as discussed
Screenshot 2022-02-24 at 6 46 09 PM
.

dattranm:
@ionicsolutions Regarding to Issue #94, I have implemented model_download function for feature recommender, as well as cache check for model path in user local. Can you review it and lmk if anything? Thank you so much!

ionicsolutions (approved):

kajanansangar (approved):

PR #98, opened 2022-02-24
Bump Spark 3.x to latest releases

ionicsolutions:


ionicsolutions:
@Vishnu-Gowthem The failure was simply due to a bug in an action (I already filed https://github.com/vemonet/setup-spark/issues/11) which led to the situation that it claimed to have installed Spark successfully when in fact nothing was installed at all.

So it looks like we're fully compatible with the latest stable Spark releases.

PR #107, opened 2022-03-01
Feature Recommender Data file locally save

dattranm:
Change the location of the preloaded csv file:
- Now it will be included inside the package, instead of taking it from github resource


kajanansangar (changes requested):
site package needs to be added into requirment.txt to @dattranm

dattranm:
> @dattranm Have you checked that the file is included in the wheel we build for PyPI? I believe the CSV needs to be added to `setup.py`, cf. https://setuptools.pypa.io/en/latest/userguide/datafiles.html

Thanks @ionicsolutions . I have added the file into setup.py

ionicsolutions (approved):
LGTM

ionicsolutions:
> site package needs to be added into requirment.txt to @dattranm

I believe we're good, that's an internal module: https://docs.python.org/3.9/library/site.html

PR #112, opened 2022-03-02
Ts databricks change

ranjanravish:
Added databricks condition in time_series scripts

### DIFF
--- a/src/main/anovos/data_ingest/ts_auto_detection.py
+++ b/src/main/anovos/data_ingest/ts_auto_detection.py
@@ -37,7 +37,6 @@ import dateutil.parser
 import pandas as pd
 import numpy as np
 
-
 ###regex based ts parser function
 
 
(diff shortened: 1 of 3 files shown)


src/main/anovos/data_ingest/ts_auto_detection.py
```diff
@@ -37,6 +37,27 @@
...
+
+    """
+    punctuations = ":"
+    for x in output_path:
+        if x in punctuations:
+            local_path = output_path.replace(x, "")
+            local_path = "/" + local_path
+    return local_path
+
+
```
VishnuGowthem, line 60: rather than repeating the same function def in 2 places, can you move it to utils and define it once and import it from utils and use ?

VishnuGowthem: Also rather than leaving the docstrings empty 
can we add meaningful examples in this case 
Like if dbfs://sample_path
is passed
then output returns 
/dbfs//sample_path ?

VishnuGowthem (changes requested):

PR #118, opened 2022-03-02
Feature Recommendation get method bugfix

dattranm:
Fix get method bug in feature_recommendation
Change the notebook accordingly for model_download to work properly there


ionicsolutions:
> @ionicsolutions Along with this, should I import the `model_download()` into feature_recommendation/feature_exploration so that users don't need to import featrec_init when the model download requirement message come out?

I don't think importing an unused method is a good idea. But we should tell users more precisely what they should do, i.e., which module they need to import and where to find the download function. You can have a look at https://www.nltk.org/data.html for an example.

Further, can you please add some basic unit tests that would catch this problem should it arise in the future? Otherwise, we'll continue flying blind here.

dattranm:
> > @ionicsolutions Along with this, should I import the `model_download()` into feature_recommendation/feature_exploration so that users don't need to import featrec_init when the model download requirement message come out?
> 
> I don't think importing an unused method is a good idea. But we should tell users more precisely what they should do, i.e., which module they need to import and where to find the download function. You can have a look at https://www.nltk.org/data.html for an example.
> 
> Further, can you please add some basic unit tests that would catch this problem should it arise in the future? Otherwise, we'll continue flying blind here.

Got it. I'm working on unit tests for all 3 sub-modules inside feature_recommender

kajanansangar (approved):
reviewed both files and can be merged

PR #119, opened 2022-03-03
Update Anovos v0.2.0 example notebooks

sinuochen:


dattranm (commented):
There is no `examples/output` folder in all branches originally on GitHub, so any notebooks that requires path to `examples/output` fail to run locally without creating the `examples/output` folder beforehand (for example, data_analyzer_ts_analyzer.ipynb)

No code logic change, so all codes are still working properly

All hyperlink in notebooks were checked and now work properly.

Some local path of got exposed in the notebook (Users/sinouchen/….), maybe consider hiding warnings or clearing the output of those cells after finishing the runs and then recommit

sinuochen:
I managed to hide most of the local paths except for a few cells in data_transformer__transformers.ipynb because some of them contain printed outputs and the messages persist after trying to remove them, or the warnings are raised by us within the function. Let me know if any changes are required :)

sourjyasen07:
Below are some of the comments from my side -

@ranjanravish 

**data_ingest__ts_auto_detection.ipynb**

1. While we're describing `ts_loop_cols_pre`, can we explain the different list elements for better clarity
2. Example 3 (varying tz_offset) has similar output owing to the data / field limitations. Can we specify that as well. 

**data_analyzer__ts_analyzer.ipynb**

1. Example 3 we're detailing based on the hourly category. But since there's one category owing to the data / field limitations (it doesn't have time stamp) it's better to call that out separately.


@sinuochen 

**data_report.ipynb**

1. Possible to add the `Time Series Analyzer` details in the notebook so that we're in sync with the details as applicable for other tabs

sourjyasen07 (commented):
changes are good to be merged

PR #127, opened 2022-03-07
Feature Recommender Unit Tests

dattranm:
Add unit tests for Feature Recommender:
-Unit test for featrec_init functions
-Unit test for feature_exploration functions
-Unit test for feature_recommendation functions
-Small typo fix for feature_recommender_demo example notebook


dattranm:
Unit tests are failing for feature_recommender because there is no download model method here. Potential solution is to include model_download() in all the unit test scripts, but that will cause the test to redownload the model everytime it starts on anywhere. @ionicsolutions Can you give us a suggestion on this? Tysm!

.github/workflows/unit.yml
```diff
@@ -61,10 +61,27 @@ jobs:
...
       - name: Run the unit tests for data_ingest, analyzer, drift modules
         run: |
           pytest -v --cov=src/main/anovos src/test/anovos/data_ingest src/test/anovos/data_analyzer src/test/anovos/drift
 
       - name: Run unit tests for transformers module
         run: |
           pytest -v --cov=src/main/anovos src/test/anovos/data_transformer
+
+      - name: Run the unit tests for feature_recommender
+        run: |
```
VishnuGowthem, line 86: As discussed, do change order to execute feature recommender above transformers

src/test/anovos/feature_recommender/test_featrec_init.py
```diff
@@ -0,0 +1,42 @@
...
+    camel_case_split,
+    recommendation_data_prep,
+)
+
+sample_data_2 = "./data/feature_recommender/test_input_fr_2.csv"
+
+
+@pytest.fixture
+def example_attr_2():
+    return pd.read_csv(sample_data_2)
```
VishnuGowthem, line 14: I think sample_data_2 can be defined inside example_attr_2() instead of a global variable.

src/test/anovos/feature_recommender/test_featrec_init.py
```diff
@@ -0,0 +1,42 @@
...
+
+
+def test_camel_case_split():
+    test_input_1 = "accountWeeks"
+    test_output_1 = camel_case_split(test_input_1)
+    assert test_output_1 == "account Weeks "
+
+    test_input_2 = "account Weeks"
+    test_output_2 = camel_case_split(test_input_2)
+    assert test_output_2 == "account Weeks "
```
VishnuGowthem, line 32: why is there a " " at the end after the split ? Is this prt of the defined functionality too ? 
Can we also add few more assert statements 
- one without any camelcase in the text
- one with multiple camelcase words in the text

src/test/anovos/feature_recommender/test_feature_exploration.py
```diff
@@ -0,0 +1,177 @@
...
+model_download()
+
+
+def test_list_all_industry():
+    test_df = list_all_industry()
+    assert len(test_df) > 0
+    assert "Industry" in test_df.columns
+    assert test_df.iloc[:, 0].nunique() == len(test_df)
+    for i in range(len(test_df)):
+        assert "," not in test_df.iloc[i, 0]
```
VishnuGowthem, line 27: Why are we checking if "," is not there in the Industry list 
What if a new industry being added has "," as part of its title?

Also Can I suggest something like checking for a couple of basic Industries if its available - like ones we feel important to have in mandatorily as a check.

src/test/anovos/feature_recommender/test_feature_exploration.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def test_list_all_usecase():
+    test_df = list_all_usecase()
+    assert len(test_df) > 0
+    assert "Usecase" in test_df.columns
+    assert test_df.iloc[:, 0].nunique() == len(test_df)
+    for i in range(len(test_df)):
+        assert "," not in test_df.iloc[i, 0]
```
VishnuGowthem, line 36: same comments as above. Suggest checking for specific use cases.

src/test/anovos/feature_recommender/test_feature_exploration.py
```diff
@@ -0,0 +1,177 @@
...
+    test_df = list_all_pair()
+    assert len(test_df) > 0
+    assert "Industry" in test_df.columns
+    assert "Usecase" in test_df.columns
+    assert test_df.groupby(["Industry", "Usecase"]).ngroups == len(test_df)
+
+
+def test_process_usecase():
+    assert process_usecase("fraud", semantic=True) == "fraud detection"
+    assert process_usecase("fraud", semantic=False) == "fraud"
```
VishnuGowthem, line 49: Can we also have some negative checks. like assert not process_usecase("fraud", semantic=False) == <something_unrelated>

src/test/anovos/feature_recommender/test_feature_exploration.py
```diff
@@ -0,0 +1,177 @@
...
+
+
+def test_process_industry():
+    assert process_industry("telco", semantic=True) == "telecommunication"
+    assert process_industry("telco", semantic=False) == "telco"
```
VishnuGowthem, line 54: same comment as above

src/test/anovos/feature_recommender/test_feature_recommendation.py
```diff
@@ -0,0 +1,231 @@
...
+    assert "AccountWeeks" in sankey_plot.data[0]["node"]["label"]
+    assert "telecommunication" not in sankey_plot.data[0]["node"]["label"]
+    assert (
+        "banking financial service and insurance"
+        not in sankey_plot.data[0]["node"]["label"]
+    )
+
+    sankey_plot_2 = sankey_visualization(
+        test_df_2, industry_included=True, usecase_included=True
+    )
```
VishnuGowthem, line 217: Is there something we can check to confirm on this method call ?

VishnuGowthem (changes requested):

src/test/anovos/feature_recommender/test_feature_recommendation.py
dattranm: For this, the type of the variable will be plot, and there are elements inside that can only be accessed while the object is plot.

src/test/anovos/feature_recommender/test_feature_exploration.py
dattranm: "," in here is to check if the input dataset used for Feature Exploration is the proper one. "," in the Industry/Usecase means that there are more than 1 Industry/Usecase in that particular cell.

> Also Can I suggest something like checking for a couple of basic Industries if its available - like ones we feel important to have in mandatorily as a check.
Sure, I will add it.

src/test/anovos/feature_recommender/test_featrec_init.py
dattranm: > why is there a " " at the end after the split ? Is this prt of the defined functionality too ?

Yeah it's part of it

> Can we also add few more assert statements

Got it

src/test/anovos/feature_recommender/test_feature_exploration.py
dattranm: Same as above

VishnuGowthem (approved):

PR #134, opened 2022-03-20
Minor Fixes @ TS Module

sourjyasen07:


dattranm:
Looks good to me. All functions run properly and there are no other bugs when running end-to-end reports

PR #144, opened 2022-04-07
phik replaced by spark spearman correlation

nisha20verma:
- Categorical columns are not supported anymore.
- Reference to popmon/phik is removed


kajanansangar:
@nisha20verma I am closing this PR since we have decided to keep Phik as well spearman

def correlation_matrix(
    spark, idf, list_of_cols="all", drop_cols=[], stats_unique={}, print_impact=False,is_numeric_colsOnly= False
):

The above function should call the respective functions accordingly. 

you can define two additional functions as follows and call accordingly:

def correlation_matrix_num_cat(....) And
def correlation_matrix_num(..)

kajanansangar:
@nisha20verma I am closing this PR since we have decided to keep Phik as well spearman

def correlation_matrix(
    spark, idf, list_of_cols="all", drop_cols=[], stats_unique={}, print_impact=False,is_numeric_colsOnly= False
):

The above function should call the respective functions accordingly. 

you can define two additional functions as follows and call accordingly:

def correlation_matrix_num_cat(....) And
def correlation_matrix_num(..)

PR #146, opened 2022-04-08
Add data_sample function and unit test

dattranm:


src/main/anovos/data_ingest/data_sampling.py
```diff
@@ -0,0 +1,68 @@
...
+def data_sample(
+    idf,
+    list_of_cols="all",
+    drop_cols=[],
+    fraction=0.1,
+    method_type="random",
+    stratified_type="proportionate",
+    seed_value=12,
+    unique_threshold=0.5,
+):
```
kajanansangar, line 14: @dattranm inline comments are missing here, please do add them and let me know

kajanansangar (changes requested):

src/main/anovos/data_ingest/data_sampling.py
```diff
@@ -0,0 +1,68 @@
...
+    if type(fraction) != float and type(fraction) != int:
+        raise TypeError("Invalid input for fraction")
+    if fraction < 0 or fraction > 1:
+        raise TypeError("Invalid input for fraction. Fraction value is between 0 and 1")
+    if type(seed_value) != int:
+        raise TypeError("Invalid input for seed_value")
+    if method_type not in ["stratified", "random"]:
+        raise TypeError("Invalid input for data_sample method_type")
+    if method_type == "stratified":
+        if stratified_type not in ["proportionate", "disproportionate"]:
```
kajanansangar, line 24: I would suggest using terms such as "population" & "balanced", "population" meaning stratified sampling outcome follows the population distribution and "balanced" meaning stratified sampling  output follows the balanced distribution based on the strata columns

kajanansangar (changes requested):

nisha20verma:
- Putting sampling function under data_ingest.py instead of a separate module
- fraction, seed, unique_threshold >> “0.5” or 0.5 - both should be ok as long as we can convert the string input into a valid numeric value.
- Check in Line 42 itself if strata col is a categorical column or not,  then Line 44 will be redundant
- Error messaging is not correct if the cardinality is higher than the given threshold.

Suggestions to make coding style consistent in all anovos functions:

- from pyspark.sql.functions import concat, lit >>> Using F.concat, F.lit (just to be consistent with other coding style)
- Using “list_of_cols” instead of “strata_cols” to be consistent with all anovos functions
- Also, we do error handling for invalid input columns in a slightly different manner and it is consistent for all other anovos functions. Maybe you can refer to that.

dattranm:
@nisha20verma Thanks for the review! I have addressed most of your comments above. For the suggestion, as we discussed, I leave the strata_cols for data_sample function to keep its meaning. I also changed the way to handle errors to somewhat standardize with what we have in anovos. Can you review it again and lmk if anything?

dattranm:
Thanks for pointing out @sinuochen. I have changed accordingly to make sure only the function check on strata_cols and such only on method_type=='stratified'

PR #148, opened 2022-04-10
bug: date column on cast as double

nisha20verma:


dattranm:
Bugs confirmed fixed, with the fix for null values as well. All datetime functions are running as expected. It's good to merge!

kajanansangar (approved):

PR #150, opened 2022-04-11
refactoring arguments

nisha20verma:


src/main/anovos/data_analyzer/validations.py
```diff
@@ -0,0 +1,187 @@
...
+            "uniqueCount_computation",
+            "mode_computation",
+            "measures_of_cardinality",
+            "IDness_detection",
+            "biasedness_detection",
+            "invalidEntries_detection",
+        ):
+            all_valid_cols = discrete_attributes(idf)
+        else:
+            all_valid_cols = num_cols + cat_cols + other_cols
```
ionicsolutions, line 49: I don't think that this approach is a good idea for three reasons:

1) It relies on `__name__`, which is a Python internal.
2) It splits up the way the actual function is defined and hides an important part from the user looking at the code. If someone renames the function or changes its arguments, it breaks this decorator.
3) This decorator is very complex, as it tries to manage the input of a large number of functions and their different arguments in just one large function.

Instead of using a decorator, I suggest to follow the principles "Explicit is better than implicit." and "Simple is better than complex." proposed in [The Zen of Python](https://peps.python.org/pep-0020/) and create a set of small, testable, easy-to-read functions that each deal with one particular argument.

For example:

```python
def parse_columns(list_of_cols: Union[str, List[str]]):
    ...

def parse_boolean(boolean_arg: Union[str, bool, int]):
   ...
```

Each of these functions should be independently tested.

Then, use the helper functions within each of Anovos' functions:

```python
def feature_engineering_method(list_of_cols, treatment, ...):

    list_of_cols = parse_columns(list_of_cols)
    treatment = parse_boolean(treatment)
   ...
```

This is easy to understand, easy to test/debug and easy to maintain.

ionicsolutions (changes requested):

src/main/anovos/data_analyzer/validations.py
ionicsolutions:
```python
@numerical_columns(arg_name="list_of_cols")
@...
@...
def feature_engineering_method(list_of_cols):
     ...
```

ionicsolutions:
Per discussion in the weekly meetings

PR #156, opened 2022-04-14
removing_autoencoder_tensorflow

nisha20verma:
Commenting out autoencoder due to incompatibility between tensorflow and M1 chip MacBook


kajanansangar:
@dattranm please help review this and test it on Apple M1 Chip Machine too

dattranm:
The branch removes Tensorflow and the function depend on it (autoencoder_latentFeature). It is able to run on M1 Mac machine with some additional installation required @kajanansangar

PR #158, opened 2022-04-20
Datapane Version Upgrade

sourjyasen07:
Script changes made to accomodate new datapane version; 
Changes to the version made specific to dependency packages;


requirements.txt
```diff
@@ -4,9 +4,9 @@ scipy<=1.8.0
 scikit-learn==1.0.2
 sympy>=1.6,<=1.9
 matplotlib>=3.4.3,<=3.5.1
-datapane==0.12.0
-jinja2<3.1.0
-plotly>=5.2.1,<=5.6.0
+datapane
```
ionicsolutions, line 7: Please pin the version to avoid issues like we just had where an upstream upgrade broke all Anovos packages on PyPI.

GitHub's Dependabot will let us know when there is a new version, then we can test and decide when to upgrade.

requirements.txt
```diff
@@ -4,9 +4,9 @@ scipy<=1.8.0
...
+jinja2
```
ionicsolutions, line 8:
```suggestion
```

`jinja2` can be removed, we are not directly depending on it.

requirements.txt
```diff
@@ -4,9 +4,9 @@ scipy<=1.8.0
...
+plotly
```
ionicsolutions, line 9:
```suggestion
```

same for `plotly`

ionicsolutions (changes requested):

ionicsolutions (approved):

PR #170, opened 2022-05-06
Add recommend_type function and unit test

dattranm:


sinuochen:
Hi @dattranm, sharing my feedback for the recommend_type function:
* If a categorical attribute originally contains null and all non-null values are integer in string format, idf_inter.distinct().na.drop().count() will be smaller than idf_inter.distinct().count() by 1 because idf_inter still contains null. Then this attribute would not be included in the final result.
* If all of rec_cols, ori_form, ori_type, rec_form and rec_type are empty, a ValueError (can not infer schema from empty dataset) is raised. I think returning an empty Dataframe would be more suitable.
* An other suggestion is that we could add another column to indicate the number of distinct values in each attribute or to state the reason of this recommendation. It may help users make decision.

We can have another discussion to see whether these comments make sense or not.

dattranm:
@sinuochen Thanks a lot for the comment. If possible, can you take a look again to see if all of your comments have been addressed? Tysm!

sinuochen:
@dattranm I have reviewed the new changes and the function is working properly. Thanks!

sinuochen (approved):

PR #182, opened 2022-05-17
adding parameter for basic_report

VishnuGowthem:


sourjyasen07:
changes have been tested out and are looking good to me

PR #185, opened 2022-05-17
Dynamically import Tensorflow based on M1 Machine status

dattranm:


kajanansangar:
@dattranm please check why code quality check is failing for this PR

ionicsolutions (approved):
I think this is fine as a workaround.

As discussed, we should aim to switch to PyTorch entirely, which we're already using for the feature recommender.

kajanansangar (approved):

PR #186, opened 2022-05-18
Move the main workflow script inside the package

### LINKED ISSUE #183: main script should be included in Anovos package
Currently, `main.py` is not included in the wheel file.

Hence, users have to navigate to the GitHub repository and retrieve the correct version of `main.py` for the version of Anovos they have installed if they want to run Anovos workloads from a configuration file. (It's not sufficient to get the latest version of `main.py`, as the version the user has installed might be incompatible with it.)

Instead, users should be able to just install Anovos from a wheel (e.g., `pip install anovos`) and run workloads, without having to download a file from GitHub or clone the GitHub repository.

ionicsolutions:
Fixes #183


VishnuGowthem (approved):
I tested branch in my local notebook as well to double check. I am good with the changes.
@ranjanravish can you help test this branch on Azure databricks so we can merge it to main.

kajanansangar (approved):

PR #187, opened 2022-05-19
Fixing the Pie Chart Issue

sourjyasen07:
The input data to Pie Chart was not being read properly which was causing it to not populate the chart labels, etc. The latest change addresses the same.

### DIFF
--- a/src/main/anovos/data_report/report_generation.py
+++ b/src/main/anovos/data_report/report_generation.py
@@ -531,8 +531,8 @@ def executive_summary_gen(
         )
         # @FIXME: never used local variable
         text_val = list(list(obj_dtls.values())[0][0].items())[8][1]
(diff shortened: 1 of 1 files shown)


kajanansangar:
@sourjyas i am getting the following error while running in my local . Refer the log below,

https://justpaste.it/4bwne

kajanansangar (changes requested):
End to end run is failing https://justpaste.it/4bwne

dattranm:
@sourjyasen07 Hey! The branch works on Intel chip Mac and doesn't work on M1 Machine. This behavior is likely inherited from Main branch. In details, on M1 machine it stops on report_preprocessing step on my test.

kajanansangar (approved):

PR #192, opened 2022-05-24
Debug and fix failing transformer tests

### LINKED ISSUE #169: Data transformer unit tests sometimes fail with StackOverflowError
see e.g. https://github.com/anovos/anovos/runs/6288850865?check_suite_focus=true#step:15:2284 and https://github.com/anovos/anovos/runs/6273324296?check_suite_focus=true#step:14:3850

It seems to work fine sometimes. Not sure whether there's a pattern.

ionicsolutions:
Debug and Fix #169


ionicsolutions:
@nisha20verma @kajanansangar @VishnuGowthem I suspect that persisting the DataFrames in the transformers modules is the cause of the issues we see with the tests.

If we run the unit tests in separate groups, they run fine as long as we limit the size of the dataframe we use for the first `matrixFactorization()` test. Once we run them all together (i.e., in direct succession), we start to see failures in seemingly arbitrary test functions in  the `transformers` module. All subsequent tests that call into Spark fail as well, with a connection timeout, indicating that the Spark instance has died and is no longer available.

We've been using just one single Spark session for all tests (this is why running the tests in batches works), which I've now reduced to using a new Spark session for each test module. Nevertheless, it seems that we have a memory leak here and should probably actively de-persist the DataFrames.

According to Git, @nisha20verma added at least some of the persists in February as part of some optimization. Let's make sure that this isn't causing issues for users of Anovos.

kajanansangar:
@ionicsolutions thanks for getting this identified. @VishnuGowthem @nisha20verma could you please review this and leave your comments on Killian's feedback.

kajanansangar (approved):

PR #193, opened 2022-05-24
Delete type check in data_ingest/write_dataset

miker2241:


miker2241:
Delete type check from data_ingest/write_dataset because we run into a ValueError before the actual type check.

PR #215, opened 2022-06-14
Fix #205 - updated version.py to also 0.3.0

### LINKED ISSUE #205: Version is defined in two places and doesn't match
https://github.com/anovos/anovos/blob/main/setup.py says 0.3.0

https://github.com/anovos/anovos/blob/main/src/main/anovos/version.py says 0.2.1

VishnuGowthem:
updated version.py file to recent release version


ionicsolutions (approved):
Can we have the version defined in one place only?

With a `setup.cfg`, it's possible to do something like

```
[metadata]
name = anovos
version = attr: anovos.__version__
```

kajanansangar:
> Can we have the version defined in one place only?
> 
> With a `setup.cfg`, it's possible to do something like
> 
> ```
> [metadata]
> name = anovos
> version = attr: anovos.__version__
> ```

@VishnuGowthem  can you please make the changes as per the Killian's comment and ask for a review? This needs to be closed by tomorrow

setup.cfg
```diff
@@ -1,2 +1,4 @@
 [metadata]
 description_file = README.md
+name = anovos
```
ionicsolutions, line 3: The `name` is also defined in the `setup.py` file. Does one file take precedence over the other?

PR #235, opened 2022-07-04
Outlier detection optimization

sinuochen:


src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -721,185 +713,283 @@ def outlier_detection(
...
+            elif detection_side == "lower":
+                return lower_v
+            else:
+                return lower_v + upper_v
+
+        return inner
 
-    odf = idf.withColumn("outliered", f_composite_outlier(*list_of_cols))
+    odf = idf
     odf.persist()
```
dattranm, line 912: In general, `persist` means that `odf` will be cached to Memory and Disk. In general, it is not recommended to do `persist` inside a function without using `unpersist`, because it can cause issues after finishing the function, as the `odf` will be on users' memory/disk even after the function finishes without their acknowledgement. I would recommend to completely remove the `persist` in here, to avoid issues in the future.

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -721,185 +713,283 @@ def outlier_detection(
...
 
-    odf = odf.drop("outliered")
+    def unionAll(dfs):
+        first, *_ = dfs
+        return first.sql_ctx.createDataFrame(
+            first.sql_ctx._sc.union([df.rdd for df in dfs]), first.schema
+        )
+
+    odf_union = unionAll(list_odf)
+    odf_agg = odf_union.groupby("attribute").pivot("outliered").count()
```
dattranm, line 953: Instead of doing it like this, as my experiment with `IV_calculation`, you should run the function(`groupby`, `agg`, etc) inside a loop for every column, then union all the results, instead of union the dataframes first, then run the function. It will help with optimization, as it will reduce the stress on the Spark Session.

An example on what I did on `IV_calculation`:
```
    list_df = []
    for col in list_of_cols:
        df_agg = (
            idf_encoded.select(col, label_col)
            .groupby(col)
            .agg() 
    . . . # All your functions inside
        list_df.append(out_df)
    odf = unionAll(list_df)
```
Reference: [IV_calculation Optimization](https://github.com/anovos/anovos/blob/f2e6cce7d6a5bbd2346b448b788e00402b06a934/src/main/anovos/data_analyzer/association_evaluator.py#L540)
You can test this out to see if there is any improvement.

src/main/anovos/data_analyzer/quality_checker.py
```diff
@@ -721,185 +713,283 @@ def outlier_detection(
...
         odf = idf
 
-    odf_print = spark.createDataFrame(
-        output_print, schema=["attribute", "lower_outliers", "upper_outliers"]
-    )
     if print_impact:
-        odf_print.show(len(list_of_cols))
+        odf_print.show(len(list_of_cols) + len(skewed_cols), False)
 
     return odf, odf_print
```
dattranm, line 994: If possible, can we make it so that it only returns `odf_print` when `print_impact==True`? As we can see in the code, `odf_print` requires a pretty significant calculation power as well, and if the user do not want to see it, there is no reason to go through the calculation everytime.

dattranm (changes requested):

src/main/anovos/data_analyzer/quality_checker.py
sinuochen: Sure, I have removed this line.

src/main/anovos/data_analyzer/quality_checker.py
sinuochen: Thanks for pointing out Dat. I have moved the aggregation inside the for loop. The execution time did improve 👍

src/main/anovos/data_analyzer/quality_checker.py
sinuochen: I agree that calculating `odf_print` is relatively time-consuming. My concern is that all functions in `quality_checker` module return two outputs: `odf` and `odf_print`. It might not be suitable if we make `odf_print` optional only for one function. Thus, I have not implemented this change. But in the future we can discuss whether to make `odf_print` optional for all functions in `quality_checker`.

dattranm: As discussed with @kajanansangar and our discussion @sinuochen, `odf_print` should be optional to calculate and will only be returned with `print_impact==True`.

dattranm (changes requested):
Change `odf_print` to be optional, only return when `print_impact==True`. Everything else looks good to me.

dattranm (approved):
All required changes have been addressed. Will approve and merge now.

PR #250, opened 2022-07-29
outlier_categories optimization

dattranm:


dattranm:
@ChandraShekhar2 Can you clean the code using `black`? It is a requirement to push the code to Anovos.

src/main/anovos/data_transformer/transformers.py
```diff
@@ -3422,6 +3421,8 @@ def outlier_categories(
     if output_mode not in ("replace", "append"):
         raise TypeError("Invalid input for output_mode")
 
+    idf = idf.persist(pyspark.StorageLevel.MEMORY_AND_DISK)
```
dattranm, line 3424: Have you tested the runtime with and without `persist`? By default, `idf` when being read in, is cached inside Spark Session memory already. This might cause some potential runtime/memory issue. Also, if you decide to put `persist`, you should `unpersist` it at the end of the function as well.

src/main/anovos/data_transformer/transformers.py
```diff
@@ -3430,39 +3431,46 @@ def outlier_categories(
...
+                    .where(~((F.col("cumu") >= coverage) & (F.col("lag_cumu") >= coverage)))
+                    .where(F.col("rank") <= (max_category - 1))
+                    .select(F.lit(i).alias("attribute"), F.col(i).alias("parameters"))
             )
             if index == 0:
                 df_model = df_cats
             else:
                 df_model = df_model.union(df_cats)
 
+    df_params = df_model.rdd.groupByKey().mapValues(list).collect()
```
dattranm, line 3459: It is the best if we can minimize the number of action (`collect`) in the code, optimally it should not be inside the function. Can you think of a way to only use transformation operation (`select`, `withColumn`, etc) to do this, without having to use `collect`.

src/main/anovos/data_transformer/transformers.py
```diff
@@ -3430,39 +3431,46 @@ def outlier_categories(
...
+    dict_params = dict(df_params)
+    broadcast_params = spark.sparkContext.broadcast(dict_params)
```
dattranm, line 3461: Have you tested the runtime with and without `broadcast`? Also the utilization of memory and resource as well. Since I think we only use 1 machine here, I'm not sure how `broadcast` can help.

dattranm (commented):
@ChandraShekhar2 I have some comments on the branch. Please note.

src/main/anovos/data_transformer/transformers.py
cshekhar17: I am not sure why you are saying that we only use 1 machine here ? 
broadcast variables are shared across all executers. 
Why Spark Broadcast is used?
When huge datasets are needed to be cached in executors, the broadcast variables are used. Let us imagine, we need to lookup for pin codes or zip codes for doing a transformation and in this case, it is neither we will be able to query the huge database every time nor it is practical to send the huge lookup for the table every time to the executors. The only solution is by converting the table lookup into a broadcast variable and will be cached by Spark in every executor in reference for future use.

Reference - https://www.educba.com/spark-broadcast/

src/main/anovos/data_transformer/transformers.py
cshekhar17: From our experience using Spark Scala - when dataset is huge and also read multiple times in the code then we should use persist (memory and disk if its very huge, only memory if its not big enough). We found significant performance improvement in our spark scala code in the past after using it. Here also we used 100 GB dataset and were using data frame multiple times for each column so persist is useful. I will test with and without persist and come back with results. I will also use unpersist in the end.

src/main/anovos/data_transformer/transformers.py
cshekhar17: Here I minimised use of collect() to only once, earlier it was being used per column. 
What do you mean by this "optimally it should not be inside the function" ?
Regarding this "Can you think of a way to only use transformation operation (select, withColumn, etc) to do this, without having to use collect", I will have to work more on this and come back.

kajanansangar (approved):

PR #255, opened 2022-08-05
Measures of cardinality performance optimisation

kajanansangar:
@dattranm creating this PR for you to merge


dattranm:
@ChandraShekhar2 Thank you for the PR. 

- When you work with Anovos, whenever you are finished with your code, can you clean it up with black? (pip install black). Let's say you change src/main/anovos/data_analyzer/association_evaluator.py, you can do black src/main/anovos/data_analyzer/association_evaluator.py before adding to your commit. The reason is Anovos we use lint and black quality standard to make sure our code has proper formats.
- Is there any optimization solution for the exact method? In here you proposed an approximate method and test the runtime of it, but there is still the original exact method for cardinality check here. Do you think you can come up with some algorithm optimization instead for it?

dattranm (approved):
Everything is good to merge now.

PR #266, opened 2022-08-11
Local execution of Anovos workloads

ionicsolutions:


ionicsolutions:
This PR is ready for review and can be merged as-is.

There are a few changes that need to be made right before and right after the next release.

**Right before the release:**

- Change the Spark version in the local workflow to 3.2.2 (right now, there is no anovos-notebook image for 3.2.2 but there will be right after the release)

**After the release has completed:**
- Change the image in run_workflow.sh to `anovos/anovos-worker-...` (we can now pull from Docker Hub)
- Adapt the README accordingly and add information how to specify the Anovos and Spark version when calling run_workflow.sh
- Add a note to the main README that advertises this way to test/explore Anovos

mathiaspet (approved):

config/configs_basic.yaml
```diff
@@ -71,12 +71,11 @@ timeseries_analyzer:
   max_days: 3600
 
 anovos_basic_report:
-  basic_report: True
+  basic_report: False
```
dattranm, line 74: This is the basic report config. Is there any reason why `basic_report` is set to False?

config/configs_basic.yaml
```diff
@@ -313,17 +312,6 @@ report_preprocessing:
...
-  master_path: 'report_stats'
-  id_col: 'ifa'
-  label_col: 'income'
-  corr_threshold: 0.4
-  iv_threshold: 0.02
-  drift_threshold_model: 0.1
-  dataDict_path: 'data/income_dataset/data_dictionary.csv'
-  metricDict_path: 'data/metric_dictionary.csv'
-  final_report_path: 'report_stats'
-
```
dattranm, line 326: This part is needed to generated the basic report as well. Any reasons why we are deleting this?

local/README.md
```diff
@@ -0,0 +1,37 @@
...
+
+Build the `anovos-worker` image:
+
+```bash
+docker build . -t anovos-worker
+```
+
+You can specify the version of Apache Spark and Anovos using build args:
+```bash
+docker build --build-arg spark_version=3.2.1 --build-arg anovos_version=0.3.0 .
```
dattranm, line 17: Do we need to update this with every anovos version for `anovos_version`? Is there anyway for this to read the version from `version.py`, aka what you did above in the local workflow:
`export ANOVOS_VERSION=cat src/main/anovos/version.py | cut -d'"' -f 2`

local/rewrite_configuration.py
```diff
@@ -0,0 +1,124 @@
...
+    if path.startswith("dbfs:") or path.startswith("s3:"):
+        raise ValueError(f"Only local paths are supported: {path}")
+    input_data_paths.append(path)
+
+
+def find_input_data_paths(sub_config):
+    for k, v in sub_config.items():
+        if k == "read_dataset":
+            file_path = v["file_path"]
+            add_path(file_path)
```
dattranm, line 45: Along with this, is there any concerns related to other data ingest functions, i.e `join_dataset`, `concatenate_dataset`? As we know, they are all in the ETL data workflow, maybe we need to rule those functions out as well?

local/rewrite_configuration.py
```diff
@@ -0,0 +1,124 @@
...
+    old_path = pathlib.Path(path).absolute()
+    relative_path = old_path.relative_to(old_parent)
+    new_path = new_parent / relative_path
+    return str(new_path)
+
+
+def rewrite_input_data_paths(sub_config):
+    for k, v in sub_config.items():
+        if k == "read_dataset":
+            v["file_path"] = rewrite_path(v["file_path"], data_directory, _DATA_DIR)
```
dattranm, line 72: The same as the ETL functions concern above

local/rewrite_configuration.py
```diff
@@ -0,0 +1,124 @@
...
+    try:
+        old_path = full_config[write_key]["file_path"]
+    except KeyError:
+        pass
+    else:
+        full_config[write_key]["file_path"] = str(_OUTPUT_DIR / pathlib.Path(old_path))
+
+        
+class OrderPreservingDumper(yaml.Dumper):
+    """cf. https://stackoverflow.com/a/52621703"""
```
dattranm, line 92: Maybe delete this in our production code

dattranm (changes requested):

config/configs_basic.yaml
ionicsolutions: Yes. The basic_report triggers the computation of a correlation matrix, which leads to a memory overflow. The version that is available on PyPI does not yet have the `skip_corr_matrix` option that was introduced to solve this exact issue. Hence, I had to turn off the report generation for the workflow to run through.

Once a new version of Anovos is released, it will have that option and we can run the basic_configs.yaml workflow as it is currently available on the main branch.

config/configs_basic.yaml
ionicsolutions: See above, it is not possible to run this with the currently available PyPI release of Anovos due to an excessive amount of memory being used.

local/README.md
ionicsolutions: There is no need to update the README. It just demonstrates how one can build the image for different versions of Spark and Anovos, which most users will never do. They will just use the image we'll provide through Docker Hub. (See my comment regarding steps to take right before and right after the release of the next Anovos version.)

local/rewrite_configuration.py
ionicsolutions: There is nothing special about these functions. All of them contain a key `read_dataset` that contains a key `file_path` which we need to adapt. (The demo workflow we are running contains both a `join_dataset` and a `concatenate_dataset` block, both of which are processed fine.)

For more information, see the documentation: https://docs.anovos.ai/using-anovos/config_file.html

local/rewrite_configuration.py
ionicsolutions: See above, it's the same pattern everywhere, see https://docs.anovos.ai/using-anovos/config_file.html

local/rewrite_configuration.py
ionicsolutions: I added it as a hint so people can go read up on why this is necessary and why/how this somewhat cryptic class works. I would leave it in here as a service for future developers.

config/configs_basic.yaml
```diff
@@ -71,12 +71,11 @@ timeseries_analyzer:
   max_days: 3600
 
 anovos_basic_report:
-  basic_report: True
+  basic_report: False
   report_args:
     id_col: ifa
     label_col: income
     event_label: '>50K'
-    skip_corr_matrix: True
```
ionicsolutions, line 79: @dattranm Once the version of PyPI support this flag, the report config can be added back on this branch as well

local/README.md
ionicsolutions: Building the images for Docker Hub is handled in the GitHub Actions workflow

config/configs_basic.yaml
dattranm: Then I would suggest us to do this as a separated PR to better document it, and then after Anovos got another release, we can revert the PR and bring it back all at once.

config/configs_basic.yaml
dattranm: The same as above, suggested to make another PR to better document it.

local/README.md
dattranm: I see. I think you can clarify it a bit more, something like: For example, for spark version 3.2.1 and anovos version 0.3.0, this is the command.

local/rewrite_configuration.py
dattranm: I think we can always put this in our documentation and/or walkthrough of anovos github workflow right? @kajanansangar Do you have any suggestion for this?

config/configs_basic.yaml
dattranm: As above, I think it's better to have a separated PR for config_basics change

local/README.md
ionicsolutions: It says "You can specify the version of Apache Spark and Anovos using build args" right above? I doubt that anyone would think that `anovos_version=0.3.0` means anything other than "use Anovos version 0.3.0", or am I missing some alternative interpretation here?

config/configs_basic.yaml
ionicsolutions: See above.

config/configs_basic.yaml
ionicsolutions: I don't follow. The issue has been fixed way back in May by introducing the `skip_corr_matrix` flag (https://github.com/anovos/anovos/commit/e0da774cd1cf0a7c4e9d0fda5c78f711e1bceb3f) and setting it to `True` everywhere (https://github.com/anovos/anovos/commit/cfeb76cd1498868219080f137514a224a02bb7f5).

Unfortunately, there hasn't been a new release since then, so the latest release that is available on PyPI does not support this flag. Hence, in order to run the workflow with this release, we need to omit the report generation.

local/rewrite_configuration.py
ionicsolutions: What exactly is your concern with the comment? Are you generally opposed to commenting on implementation details or do you want to omit the link to SO and would rather have a description right here in the code?

It has also nothing to do with any GitHub workflows, it's just a way to ensure that the YAML file written to disk has the keys in the same order as the original file. While YAML as a file format is indifferent to key order, Anovos relies on the order of keys in the file to determine the order in which operations are performed.

config/configs_basic.yaml
ionicsolutions: There is no need to change this, the issue has been fixed months ago. We just don't have a way to install a version that include this fix from PyPI :-)

ionicsolutions: If you want to, you can build the Docker image for the latest main and verify that the current basic_configs.yaml runs fine inside the anovos-worker container.

local/rewrite_configuration.py
dattranm: Thanks for the explanation @ionicsolutions . This is resolved and closed for now.

local/README.md
dattranm: Got it. Will mark this as resolved.

config/configs_basic.yaml
dattranm: Got it. So I would take it that, after this PR is merged and we release 0.4.0 on PyPI, we then can open a new PR to change this back to `True` and bring back all the `skip_corr_matrix` and `report_generation` in configs_basic.yaml right?

ionicsolutions: Either that, or we wait, or live with the fact that the local_execution workflow is broken on `main`. I'd prefer to wait until right before the release but ultimately it doesn't really matter. Your choice :-)

dattranm: I think it's better to wait right before the release as well. @kajanansangar FYI

dattranm (approved):

PR #268, opened 2022-08-12
Add additional features to read_dataset and proper changes in workflow

dattranm:


src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -13,17 +13,31 @@
...
 
 
-def read_dataset(spark, file_path, file_type, file_configs={}):
+def read_dataset(
+    spark,
+    file_path,
+    file_type,
+    file_configs={},
+    parquet_conversion=False,
+    intermediate_path="",
```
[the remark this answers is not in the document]
dattranm: Resolved with [this commit](https://github.com/anovos/anovos/pull/268/commits/4efc62c98887186cebe7f164866d773f31f680e7)

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
 
     Returns
     -------
     DataFrame
 
     """
     odf = spark.read.format(file_type).options(**file_configs).load(file_path)
+    if parquet_conversion:
+        if not parquet_output_path:
+            raise TypeError(
```
ionicsolutions, line 97: Semantically, this should be a `ValueError`.

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+                "parquet_output_path cannot be blank for parquet conversion"
+            )
+    if treatment:
+        if id_cols:
+            for col in id_cols:
+                if col not in odf.columns:
+                    raise TypeError("Invalid input for id_cols: " + col)
```
ionicsolutions, line 104: This is not a `TypeError` but a `ValueError`. You should also mention explicitly mention the problem: The column is not in the dataframe. ("Invalid" can mean a lot of things.)

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+        if type(threshold_num) != int:
+            raise TypeError("Invalid input for threshold_num")
```
ionicsolutions, line 106: Type checks should use ìsinstance`, e.g., `if not isinstance(threshold_num, int):` 

See https://switowski.com/blog/type-vs-isinstance for a nice write-up on the topic.

Also, be more specific and tell the user precisely what is wrong.

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+        if type(threshold_ratio) != float or threshold_ratio < 0 or threshold_ratio > 1:
+            raise TypeError("Invalid input for threshold_ratio")
```
ionicsolutions, line 108: This mixes two types of errors (out of range and a type mismatch). Also, the error message should be more specific: "Invalid" can mean a lot of things and the user would have to look at the source code to understand what they did wrong. Tell them precisely what is wrong, e.g., that the number they gave is out of range.

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+        if (
+            type(threshold_string) != float
+            or threshold_string < 0
+            or threshold_string > 1
+        ):
+            raise TypeError("Invalid input for threshold_string")
```
ionicsolutions, line 114: As above: Use `isinstance` and be more specific in the error message.

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+        list_of_cols = list(c for c in odf.columns if c not in id_cols)
+        odf = odf.persist(pyspark.StorageLevel.MEMORY_AND_DISK)
```
ionicsolutions, line 116: Does it make sense to prescribe this for all users? If I'm running on machines with a lot of memory, I likely don't want to waste time by persisting to disk?

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+        odf_recast = odf.select(
+            [F.col(c).cast(T.DoubleType()).alias(c) for c in list_of_cols]
+        )
+        funs_count = [F.count]
```
ionicsolutions, line 120: In general, I think that the variable names could be even more clear, like `num_of_distinct_functions`

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+
+        list_count = odf.groupby().agg(*exprs_count).rdd.flatMap(lambda x: x).collect()
+
+        funs_distinct = [F.countDistinct]
+        exprs_distinct = [f(F.col(c)) for f in funs_distinct for c in list_of_cols]
+        list_distinct_count = (
+            odf.select(*exprs_distinct).rdd.flatMap(lambda x: x).collect()
+        )
+        list_ratio = [i / j for i, j in zip(list_distinct_count, list_count)]
+        list_cat_ratio = [i / j for i, j in zip(list_count_recast, list_count)]
```
ionicsolutions, line 134: Just by looking at the code, I'm not sure I understand what is computed and especially *why* it is computed here. This looks like it should be refactored into a clearly named function?

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+
+        list_schema = []
+        if id_cols:
+            for col in id_cols:
+                list_schema.append(T.StructField(str(col), T.StringType(), True))
+        for k in range(0, len(list_count)):
```
ionicsolutions, line 140: What exactly is going on here? It looks like you're guessing the appropriate type based on the number of distinct values in a column?

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+                    T.StructField(str(list_of_cols[k]), T.StringType(), True)
+                )
+
+            else:
+                list_schema.append(
+                    T.StructField(str(list_of_cols[k]), T.DoubleType(), True)
+                )
+        odf.unpersist()
+        full_schema = T.StructType(list_schema)
+        odf = (
```
ionicsolutions, line 168: Why are you reading the dataframe again?

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +55,129 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+            .schema(full_schema)
+            .load(file_path)
+        )
+    if parquet_conversion:
+        file_path = parquet_output_path
+        file_type = "parquet"
+        odf.write.format(file_type).options(**file_configs).save(
+            file_path, mode="overwrite"
+        )
+        odf = spark.read.format(file_type).options(**file_configs).load(file_path)
```
ionicsolutions, line 180: Why are you reading the dataframe for a (potentially) third time?

ionicsolutions (changes requested):

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/b3044847cfb728a4b4bec2d105ea7253e325643e

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/b3044847cfb728a4b4bec2d105ea7253e325643e

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/b3044847cfb728a4b4bec2d105ea7253e325643e

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/b3044847cfb728a4b4bec2d105ea7253e325643e

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/b3044847cfb728a4b4bec2d105ea7253e325643e

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/b3044847cfb728a4b4bec2d105ea7253e325643e

src/main/anovos/data_ingest/data_ingest.py
dattranm: The idea is to write the file in parquet_output_path in the form of parquet file, and reread it again from there.

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: To me, this kind of heuristic seems quite dangerous and incredibly difficult to debug. Why don't you have the user specify the schema and/or give them the option to override the type for certain columns?

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: Can't you just update the dataframe in-memory? Writing to a file and then reading it again seems to be a performance bottleneck.

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: Does the resulting dataframe differ, though? Shouldn't it be the same data irrespective of the file type it was read from?

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +57,188 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+        if not isinstance(threshold_ratio, float):
+            raise TypeError("threshold_ratio must be float")
+        if threshold_ratio < 0 or threshold_ratio > 1:
+            raise ValueError("threshold_ratio value must be between 0 and 1")
+        if not isinstance(threshold_string, float):
+            raise TypeError("threshold_string must be float")
+        if threshold_string < 0 or threshold_string > 1:
+            raise ValueError("threshold_string value must be between 0 and 1")
+        if not isinstance(persist_use, bool):
+            raise TypeError("persist_use must be boolean")
```
ionicsolutions, line 133: This kind of type checking is unnecessary and just makes the code longer and harder to read. If you add type hints, any modern IDE will alert users that they're using the wrong types. Internally you can add a tool like `mypy` to check that Anovos internally passes the proper types.

src/main/anovos/data_ingest/data_ingest.py
```diff
@@ -41,13 +57,188 @@ def read_dataset(spark, file_path, file_type, file_configs={}):
...
+            elif persist_option == "MEMORY_ONLY_SER":
+                odf = odf.persist(pyspark.StorageLevel.MEMORY_ONLY_SER)
+            elif persist_option == "MEMORY_ONLY_SER_2":
+                odf = odf.persist(pyspark.StorageLevel.MEMORY_ONLY_SER_2)
+            elif persist_option == "OFF_HEAP":
+                odf = odf.persist(pyspark.StorageLevel.OFF_HEAP)
+            else:
+                raise TypeError(
+                    "Input persist_option does not exist in Spark StorageLevel"
+                )
```
ionicsolutions, line 158: Why do you have the user pass a string instead of the actual storage level? You could just pass `persist_option=pyspark.StorageLevel.MEMORY_AND_DISK` instead of translating between an error-prone string and the actual StorageLevel.

src/main/anovos/data_ingest/data_ingest.py
dattranm: The general idea that we are trying to make here is to introduce a somewhat intelligent function to automatically check   and treat the schema of any DataFrame. We can also do as you suggest, but I think it would defeat the purpose of developing some intelligent logic for the treatment.

src/main/anovos/data_ingest/data_ingest.py
dattranm: We maintain this kind of format as of now for all Anovos functions. I think it would be wise to keep it consistent with other functions, and then we can make this a suggestion/issue and create a separated PR to fix those.

src/main/anovos/data_ingest/data_ingest.py
dattranm: As we test it out, by writing the dataframe out in parquet format, and reread the parquet dataframe again, the dataframe performs a lot better in terms of runtime compared with csv. At some cases, we see up to 50% at runtime reduction in parquet compared with csv input. The reason for this is because, Spark read and save parquet file in a different manner than csv, and parquet is optimized for big data size. As a result, that makes a huge improvement in terms of optimization.

src/main/anovos/data_ingest/data_ingest.py
dattranm: Resolved with https://github.com/anovos/anovos/pull/268/commits/dd553da3d2fc2ba0207a6e08d993a481743525ca

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: Does the in-memory structure of a dataframe depend on the file format it was read from? Where exactly do you see the speed-up? While reading/writing the file or while processing data in-memory?

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: I'm not convinced. This kind of heuristic is very brittle and will have to be re-tuned for every new dataset, and for many datasets you won't find thresholds that yield the correct type for every single column. Just imagine the issues that arise when you're using Anovos to process a daily batch of new data and then one day the statistics change slightly, silently causing the column types to change.

Instead of trying to guess/determine the "correct" thresholds, it will be much easier and - most importantly - more reliable for users to just write the desired schema.

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: Do you have any data/indication that lazy evaluation of `cast` causes a performance hit? I would be surprised to learn that re-casting in-memory is generally slower than writing and reading from disk.

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: I think it's an anti-pattern to "type-check" in this way. For example, checking for `bool` restricts the users and takes away a core feature of Python. There is nothing wrong with the following:

```python
some_list = []

read_dataset(..., persist_use=some_list, ...)

if persist_use:  # will evaluate to "false" if list is empty, otherwise "true"
    ...
```

But `isinstance(some_list, bool)` is `False`. You can do the same with most built-in types in Python like `str`, `int` etc. and there are many cases where this is fine and very convenient.

Same goes for passing an `int` instead of a `float`. There are very few cases where this is problematic, but if you have a check like `isinstance(some_integer, float)` in place you force the user to tediously do a "type-conversion" themselves, instead of having Python handle this automatically as it was intentionally designed to do.

src/main/anovos/data_ingest/data_ingest.py
dattranm: I think we already have a function exactly for that, `recast_column` (https://github.com/anovos/anovos/blob/a94b847408143f430b40bc8b0abc35fa1521dea5/src/main/anovos/data_ingest/data_ingest.py#L322). It allows the user to make change to the schema and cast columns as they want. It would be duplicate to do that here.

src/main/anovos/data_ingest/data_ingest.py
dattranm: `cast` transformation will definitely run in an instant after you do that operation. However, the actual runtime of this transformation will happen outside the scope of this function, aka whenever an action operation is called. This make the downstream application that use this output much more unpredictable, since in addition to the action operation runtime, some intermediate memory will be created on the spark session, and it will add additional stress on it. Instead, if rereading from the source, the runtime in general will be the same with the total runtime of `cast` and action operation, and there will be no additional stress on spark session, as well as downstream application, which is the purpose of `read_dataset`.

src/main/anovos/data_ingest/data_ingest.py
dattranm: The speed-up we see is when the output of this function, `odf`, is being used as an input in other Anovos functions. So I would say it will be in processing data in-memory down the line.

src/main/anovos/data_ingest/data_ingest.py
ionicsolutions: Then why do you need the functionality proposed in this PR in the first place? :-)

PR #274, opened 2022-08-15
av-499: optimize duplicate_detection, assign .count() to variable and…

mwjinjin:
… reuse the variable

### DIFF
--- a/src/main/anovos/data_analyzer/quality_checker.py
+++ b/src/main/anovos/data_analyzer/quality_checker.py
@@ -47,7 +47,7 @@ from anovos.shared.utils import (
 
 
 def duplicate_detection(
-    spark, idf, list_of_cols="all", drop_cols=[], treatment=False, print_impact=False
+    spark, idf, list_of_cols="all", drop_cols=[], treatment=True, print_impact=False
 ):
     """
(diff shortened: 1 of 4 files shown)


kajanansangar:
@dattranm  please help review this

dattranm (changes requested):
Thank you for the PR @mwjinjin . 
- The `count` method in general is an action operation. Have you looked into using `groupby.count()` instead? The logic might change a bit, but `groupby.count()` is a transformation operation so it might further optimize the function without the need to collect the end result. 
- In this function, `odf_print` is being produced and returned regardless of `print_impact` status. As I proposed in #235 , we should produce and return `odf_print` only when `print_impact==True`. This will significantly improve the runtime, as not every time `odf_print` is needed for the user. @sinuochen has a similar PR about this ( #235 ), can you take a look and make this change here as well?

mwjinjin:
Hi @dattranm , I committed the changes. The performance of `groupby(list_of_cols).count().drop("count")` is better than `drop_duplicates`, I tested with workflow on 50GB data, `groupby` took 479 sec while `drop_duplicates` took 532 sec.

dattranm (approved):
All the comments have been addressed. The changes look all good now and PR is good to merge.

PR #275, opened 2022-08-16
Optimize cat_to_num_supervised

sinuochen:


src/main/anovos/data_transformer/transformers.py
```diff
@@ -798,9 +802,22 @@ def cat_to_num_supervised(
...
+    )
+
+    if model_path == "NA":
+        skip_if_error = True
+        model_path = "intermediate_data"
+    else:
+        skip_if_error = False
+    save_model = True
 
+    if intermediate_data_caching:
```
kajanansangar, line 819: better change the variable name to `if persists:`

kajanansangar (approved):

src/main/anovos/data_transformer/transformers.py
```diff
@@ -769,6 +768,11 @@ def cat_to_num_supervised(
         "local", "emr", "databricks", "ak8s" (Default value = "local")
     auth_key
         Option to pass an authorization key to write to filesystems. Currently applicable only for ak8s run_type. Default value is kept as "NA"
+    intermediate_data_caching
```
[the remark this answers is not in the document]
dattranm: @sinuochen For this, you can refer to read_dataset on read_dataset_change branch, `persist_use` and `persist_option`. Reference: https://github.com/anovos/anovos/blob/dd553da3d2fc2ba0207a6e08d993a481743525ca/src/main/anovos/data_ingest/data_ingest.py#L29

src/main/anovos/data_transformer/transformers.py
```diff
@@ -824,48 +838,55 @@ def cat_to_num_supervised(
...
+                            + i
+                            + "' faced an error."
+                        )
+                        save_model = False
+                    else:
+                        raise error
 
         if df_tmp.count() > 1:
-            odf_partial = odf_partial.join(df_tmp, i, "left_outer")
+            odf = odf.join(F.broadcast(df_tmp), i, "left_outer")
```
dattranm, line 869: Using `broadcast` to optimize the join method is a nice approach, but need to also take into account the fact that 
`spark.sql.autoBroadcastJoinThreshold` by default is only set at 10MB, and the maximum size of a broadcast table is 8GB. Since both of these thresholds we cannot change nor modify, using `broadcast` to join in the function might cause some scalability issues later, so I would recommend against it.

src/main/anovos/data_transformer/transformers.py
```diff
@@ -824,48 +838,55 @@ def cat_to_num_supervised(
...
         else:
-            odf_partial = odf_partial.crossJoin(df_tmp)
-
-    odf = idf_id.join(odf_partial.drop(*list_of_cols), "tempID", "left_outer").drop(
-        "tempID"
-    )
+            odf = odf.crossJoin(F.broadcast(df_tmp))
```
dattranm, line 871: The same as the above comment, against using `broadcast` during join.

dattranm (changes requested):

sinuochen:
Thanks for your review @kajanansangar and @dattranm. I have addressed them in the latest commit. 

For now I used `persist` as the new argument name - @dattranm should we finalise a name: `persist` or `persist_use` so that it's consistent in both cat_to_num_supervised and read_dataset?

dattranm (approved):
Thank @sinuochen . Will merge and close this

PR #278, opened 2022-08-22
optimize measures_of_centralTendency

mwjinjin:


src/main/anovos/data_analyzer/stats_generator.py
```diff
@@ -415,9 +414,11 @@ def unionAll(dfs):
         )
 
     odf = unionAll(list_df)
+
     if print_impact:
+        odf.persist()
```
dattranm, line 419: It is a good practice to always do `unpersist` after `persist`. Can you add that in here?

src/main/anovos/data_analyzer/stats_generator.py
```diff
@@ -510,8 +521,9 @@ def measures_of_centralTendency(
     )
 
     if print_impact:
+        odf.persist()
         odf.show(len(list_of_cols))
```
dattranm, line 525: The same as above. Need to add `unpersist`.

dattranm (changes requested):

mwjinjin:
Hi @dattranm , I did several testing on persist() and unpersist(), pls refer to the result here: https://mobilewalla.atlassian.net/browse/AV-505?focusedCommentId=10517. Based on the testing result, doing persist() without unpersist() gives the best performance. Is it acceptable that we keep persist() without unpersist()?

dattranm:
@mwjinjin It's not always about the runtime here. Whenever you do `persist` on a DataFrame, that DataFrame will stay inside the `persist` option for as long as the Spark Session is still active, thus, will cause memory stress on the Spark session even after this whole function finishes. It is always a good practice to do `unpersist` in order to relieve the memory stress caused by `persist`, else the function/Spark Session will become much more unpredictable and unstable after.

mwjinjin:
Hi @dattranm , I just removed persist() since I think persist() does not do much help for the performance, and I agree with you that unpersisting is a good practice, thank you!

dattranm:
Thank @mwjinjin . Will approve and merge once all the tests pass.

dattranm (approved):

PR #291, opened 2022-09-08
Unique count computation rsd value correction

cshekhar17:


dattranm (changes requested):
This `rsd` change is fine. However, can you use this PR to add the value check for `rsd` as well? For example, check type, and value, and such. You can refer to https://github.com/anovos/anovos/blob/dd553da3d2fc2ba0207a6e08d993a481743525ca/src/main/anovos/data_ingest/data_ingest.py#L128 for example.

dattranm (changes requested):
The `rsd` default value you leave it as `None`, so when it is compared with 0, it gives an error in the unit test, as `None` cannot be compared with 0, `NoneType` v. `int`. You will need to exclude the `None` when doing the input check.

Also, does `rsd` have any upper limit value?

dattranm (approved):
Every comment and concern has been addressed. This PR is good to merge now. @kajanansangar can you have a final look and merge?

kajanansangar (approved):

PR #296, opened 2022-09-12
Update central tendency

ZhuLi99:
Updated the central Tendency Function to handle Example 5 in [this notebook](https://github.com/anovos/anovos/blob/main/examples/notebooks/data_analyzer__stats_generator.ipynb)

Example 5 is only using categorical columns, so the **summary_lst** values will be none. If we stick to original schema code as below:
> summary_df = spark.createDataFrame(
>         summary_lst,
>         schema=("key", "mean", "50%", "count"),
>     )

It will raise error as the values are None thus unable to determine the schema. My suggestion is to define the schema as my commit to handle None cases. Please kindly let me know if this is ok.. Thanks


VishnuGowthem (approved):

src/main/anovos/data_analyzer/stats_generator.py
```diff
@@ -493,7 +493,14 @@ def measures_of_centralTendency(
         summary_lst.append(summary_col)
     summary_df = spark.createDataFrame(
         summary_lst,
-        schema=("key", "mean", "50%", "count"),
+        schema=T.StructType(
+            [
+                T.StructField("key", T.StringType(), True),
+                T.StructField("mean", T.DoubleType(), True),
+                T.StructField("50%", T.DoubleType(), True),
+                T.StructField("count", T.LongType(), True),
```
dattranm, line 501: This change is reasonable to fix the bug you observed. However, strictly assign the type for `key`, `mean`, `50%`, and `count` will cause some format error. Refer to https://github.com/anovos/anovos/actions/runs/3038577821/jobs/4893029204#step:11:187 for the error log.

src/main/anovos/data_analyzer/stats_generator.py
```diff
@@ -493,7 +493,14 @@ def measures_of_centralTendency(
         summary_lst.append(summary_col)
```
dattranm, line 493: In here, what I would recommend is, before appending to `summary_lst`, make sure `mean`, `50%` and `count` follows the same datatype as you mentioned below. Thus, we will need to `cast` the data before appending to `summary_lst`.

dattranm (changes requested):
Thank for the Pull Request @ZhuLi99 . Here I got some comments about the change. One other thing is, please run the unit test in your local before pushing, to make sure everything work locally first.

dattranm (approved):
Thank @ZhuLi99 . All the comments have been addressed.

PR #190, opened 2022-05-24
WIP: Drift stability improvement

sinuochen:

PR #306, opened 2022-09-26
Drift function optimization

dattranm:


src/main/anovos/drift/detector.py
```diff
@@ -187,55 +254,113 @@ def statistics(
         else:
             x = (
                 source_bin.groupBy(i)
-                .agg((F.count(i) / idf_source.count()).alias("p"))
+                .agg((F.count(i) / count_source).alias("p"))
```
sinuochen, line 257: Seems that `count_source` is not defined if `sample_use` = False

src/main/anovos/drift/detector.py
```diff
@@ -187,55 +254,113 @@ def statistics(
...
+                    header=True,
+                    mode="overwrite",
+                )
 
-        y = (
-            target_bin.groupBy(i)
-            .agg((F.count(i) / idf_target.count()).alias("q"))
-            .fillna(-1)
-        )
+        y = target_bin.groupBy(i).agg((F.count(i) / count_target).alias("q")).fillna(-1)
```
sinuochen, line 267: Seems that `count_target` is not defined if `sample_use` = False

src/main/anovos/drift/detector.py
```diff
@@ -187,55 +254,113 @@ def statistics(
...
+        return first.sql_ctx.createDataFrame(
+            first.sql_ctx._sc.union([df.rdd for df in dfs]), first.schema
         )
-        .select(["attribute"] + method_type + ["flagged"])
-        .orderBy(F.desc("flagged"))
+
+    odf_union = unionAll(temp_list)
+    odf = odf_union.withColumn(
+        "flagged",
+        F.coalesce(
```
sinuochen, line 358: `F.coalesce` returns the first column that is not null so if the value can only be 0 or 1, the first value will always be returned.

src/test/anovos/drift/test_statistics.py
```diff
@@ -0,0 +1,46 @@
+import pandas
```
sinuochen, line 1: Can we name this script based on the name of the original script :) For example the unit test of detector.py  would be test_detector.py.

sinuochen (changes requested):
Hi @dattranm, I have added my review. An additional comment is that previously we changed the folder structure and the function name in PR https://github.com/anovos/anovos/pull/190. @kajanansangar can you help take a look and confirm whether we also need the following changes in this branch?
* Folder structure: main/anovos/drift_stability/
    * drift_detector.py: Containing functions related drift computation. For example, drift_statistics.
    * stability.py: Containing functions related to stability computation. For example, stability_index_computation and feature_stability_estimation

src/main/anovos/drift/detector.py
```diff
@@ -146,6 +187,32 @@ def statistics(
...
+                strata_cols=strata_cols,
+                fraction=sample_size / max(count_target, count_source),
+                method_type=sample_method,
+                stratified_type=stratified_type,
+                seed_value=sample_seed,
+            )
+            idf_source = data_sample(
+                idf_source,
+                strata_cols=strata_cols,
+                fraction=sample_size / max(count_target, count_source),
```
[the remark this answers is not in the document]
dattranm: For this, what I think and figure is that, both dataset should be sampled with the same method to ensure there is no loss of generality. For example, if the source dataset has some partial similarity with the target dataset, and only the source dataset got sampled, the similarity can completely disappear, thus cause the loss of generality here.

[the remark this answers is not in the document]
dattranm: Addressed in my newest commit.

sinuochen:
@dattranm thanks for addressing the comments. I only have one more feedback on our discussion above and I have commented the details here: https://github.com/anovos/anovos/pull/306#discussion_r1004340753.

src/main/anovos/drift/detector.py
dattranm: Thank @sinuochen . I have addressed it in https://github.com/anovos/anovos/pull/306/commits/f5b6c7b9a9219aee958509a853b2476e2116c953. Can you have a look and see if this is resolved or not.

dattranm:
Thank @sinuochen . I have addressed all of your comment and review. Can you have a final look to see if it is good enough to be approved?

sinuochen (approved):

sinuochen:
Thanks @dattranm. I have reviewed the new changes and approved the PR.

kajanansangar (changes requested):

kajanansangar (approved):

PR #307, opened 2022-10-01
Reset mistakenly overwritten config & output report when running workloads locally in Docker container

ionicsolutions:
Parts of this file were mistakenly deleted when PR #266 was squash-merged. This PR resets it to the state of #254

### DIFF
--- a/config/configs_basic.yaml
+++ b/config/configs_basic.yaml
@@ -393,14 +393,6 @@ transformers:
       imputation: True
 
 
-    # autoencoder_latentFeatures:
-    #   list_of_cols: all
-    #   reduction_params: 0.5
-    #   sample_size: 10000
-    #   epochs: 20
-    #   batch_size: 256
-
-
 write_intermediate:
   file_path: "intermediate_data"
   file_type: csv
--- a/config/configs_basic.yaml
+++ b/config/configs_basic.yaml
@@ -71,7 +71,7 @@ timeseries_analyzer:
   max_days: 3600
 
 anovos_basic_report:
-  basic_report: False
+  basic_report: True
   report_args:
     id_col: ifa
     label_col: income
--- a/config/configs_basic.yaml
+++ b/config/configs_basic.yaml
@@ -66,7 +66,7 @@ timeseries_analyzer:
   auto_detection: True
   id_col: 'ifa'
   tz_offset: 'local'
-  inspection: False
+  inspection: True
   analysis_level: 'daily'
   max_days: 3600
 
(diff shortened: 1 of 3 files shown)


config/configs_basic.yaml
```diff
@@ -66,7 +66,7 @@ timeseries_analyzer:
   auto_detection: True
   id_col: 'ifa'
   tz_offset: 'local'
-  inspection: False
+  inspection: True
```
VishnuGowthem, line 69: inspection can remain "False" in configs_basic.yaml
in PR #254 also it is False.
in configs_full.yaml it is set to True

VishnuGowthem (commented):
Other changes are good to proceed

config/configs_basic.yaml
ionicsolutions, line 69:
```suggestion
  inspection: False
```

ionicsolutions:
@VishnuGowthem Setting `inspection` to `False` causes an error in both the `demo` and the `local` workflows:

```shell
FileNotFoundError: [Errno 2] No such file or directory: '/output/report_stats/stats_dt_1_1.csv'
```

It seems that the report generation is searching for some output that is created when `inspection` is `True`?

VishnuGowthem:
> @VishnuGowthem Setting `inspection` to `False` causes an error in both the `demo` and the `local` workflows:
> 
> ```shell
> FileNotFoundError: [Errno 2] No such file or directory: '/output/report_stats/stats_dt_1_1.csv'
> ```
> 
> It seems that the report generation is searching for some output that is created when `inspection` is `True`?

@ionicsolutions  Yes seems like the case, I will look into this today.

VishnuGowthem:
> > @VishnuGowthem Setting `inspection` to `False` causes an error in both the `demo` and the `local` workflows:
> > ```shell
> > FileNotFoundError: [Errno 2] No such file or directory: '/output/report_stats/stats_dt_1_1.csv'
> > ```
> > 
> > 
> >     
> >       
> >     
> > 
> >       
> >     
> > 
> >     
> >   
> > It seems that the report generation is searching for some output that is created when `inspection` is `True`?
> 
> @ionicsolutions Yes seems like the case, I will look into this today.

Basic report was not set to True in configs_basic.yaml and I verified with True in local successfully, hence made the same change here. 

Seems when Inspection is set to False, 
the full demo run fails (will check further and raise an issue on this if required to investigate it further) 
and Basic Report runs fine. 

And when Inspection is set to True, 
both runs passes,
but for Basic report as inspection step is not a need necessity and it would run slightly longer so we can keep it for false for now. Thanks !

ionicsolutions:
@VishnuGowthem Thanks for investigating! Let's get this PR merged?

PR #318, opened 2022-10-12
stability_index_computation optimization

dattranm:


sinuochen (changes requested):
@dattranm I have left my comments above. Can you help make the changes accordingly if you think the comments make sense? Also just a small reminder - we need to update the folder name in this branch as well before merging :).

sinuochen:
Thanks @dattranm, just one more comment: I realised the only difference between line 244-253 and 255-264 is one string ("Binary" or "Numerical"), can we make that string a variable and remove the duplicate code? Something like:
```
if i in binary_cols:
    col_type = "Binary"
else:
    col_type = "Numerical"
```
Same thing happens to line 266-294 and line 296-324. Can we do the same changes?

I'm good with the other new changes, thanks.

dattranm:
Thanks @sinuochen . I have addressed the comment. Can you have a final look later?

sinuochen (approved):

sinuochen:
Thanks @dattranm. I have reviewed the new changes in the function but seems that the unit test workflow needs to be updated.

dattranm:
Thanks @sinuochen . The unit test issue will be resolved once #306 is merged, since it is related to the change and addition of module.

kajanansangar (approved):

kajanansangar (approved):

PR #323, opened 2022-10-19
select num_cols only before initiating VarClusHi

ZhuLi99:
Fixed the bug in variable_clustering function by only selecting num_cols before initiating VarClusHi object.


VishnuGowthem (approved):

VishnuGowthem (approved):

VishnuGowthem:
Going ahead with merging as all tests and demo has passed.

PR #330, opened 2022-10-28
Debug variable clustering for null

ZhuLi99:
Renamed the branch


dattranm (approved):
Thanks @ZhuLi99 for the PR. All changes have been addressed, and the PR is good to be merged and approved.

ZhuLi99:
Hi @kajanansangar @dattranm I have updated the code for variable_clustering and relevant lines in anovos_basic_report to remove passing and filtering on "stats_unique". Please let me know if anything

src/main/anovos/data_analyzer/association_evaluator.py
```diff
@@ -223,23 +217,14 @@ def variable_clustering(
 
     idf_sample = idf.sample(False, min(1.0, float(sample_size) / idf.count()), 0)
```
dattranm, line 218: As discussed, in here, can you replace `sample` with our `data_sample` function from anovos.data_ingest.data_sampling module. An example on how to implement it you can see in here:
https://github.com/anovos/anovos/blob/3d4c3d08ec4b868e182b4543831644a7a0ff2e3c/src/main/anovos/data_analyzer/association_evaluator.py#L109

dattranm (changes requested):

src/main/anovos/data_analyzer/association_evaluator.py
```diff
@@ -187,11 +186,6 @@ def variable_clustering(
     sample_size
         Maximum sample size (in terms of number of rows) taken for the computation.
         Sample dataset is extracted using random sampling. (Default value = 100000)
-    stats_unique
```
dattranm, line 190: By removing stat_unique, you will also need to remove from the workflow.py for `variable_clustering` as well. Remove `stat_unique` in here should solve the issue in workflow test: https://github.com/anovos/anovos/blob/3d4c3d08ec4b868e182b4543831644a7a0ff2e3c/src/main/anovos/workflow.py#L77

dattranm (changes requested):

ZhuLi99:
Hi @ionicsolutions , we face some errors in local run when merging from "main" to "debug-variable-clustering" branch. @Vishnu-Gowthem and I had a discussion and found out that the local run was installing Anovos 0.3.0 [here](https://github.com/anovos/anovos/blob/debug-variable-clustering-for-null/local/Dockerfile). Since our latest changes are not released yet, the local run triggered would face errors in variable_clustering due as the dataframe includes categorical columns (but in our fix, we have dropped cat_cols). Could you kindly suggest how shall we edit the local-run dockerfile to include the latest versions? Or perhaps can we skip the run-local-demo in merging? Thanks!

ionicsolutions:
> Hi @ionicsolutions , we face some errors in local run when merging from "main" to "debug-variable-clustering" branch. @Vishnu-Gowthem and I had a discussion and found out that the local run was installing Anovos 0.3.0 [here](https://github.com/anovos/anovos/blob/debug-variable-clustering-for-null/local/Dockerfile). Since our latest changes are not released yet, the local run triggered would face errors in variable_clustering due as the dataframe includes categorical columns (but in our fix, we have dropped cat_cols). Could you kindly suggest how shall we edit the local-run dockerfile to include the latest versions? Or perhaps can we skip the run-local-demo when merging from main to branch? Thanks!

We currently do not test the local execution with the latest dev versions of Anovos because there is really no point in doing so: If an Anovos workload can be executed on a VM, it will also run inside a Docker container, unless the way that we translate the workload configuration and set up the container is wrong. We've very consciously not given users the option to build an `anovos-worker` image for non-released versions of Anovos.

Hence, we should just remove the trigger on the `configs_basic.yaml` file from the GitHub Actions workflow and the problem you encountered won't happen anymore. The GitHub Actions workflow for the Docker-based local execution should only be triggered when changes are made to that feature.

kajanansangar (approved):

dattranm (approved):

PR #341, opened 2022-11-11
Drift module notebook change

dattranm:


sinuochen:
Thanks @dattranm, I have reviewed the updated notebook and I have one comment - because we changed the name of the sub-module, should we also change the name of the notebook? For example drift_stability.ipynb if we will keep one notebook for drift and stability functions.

dattranm:
Thanks @sinuochen for the comment! I have changed the notebook name and rerun it successfully. For the data_report and use_case_demo notebook, it will be changed in https://github.com/anovos/anovos/pull/359. I have restored both of them to avoid the duplication and conflict. Can you have a final look to see if this is good to merge?

sinuochen (approved):

Issue #356, opened 2022-11-28
Drop support for Python 3.7


ionicsolutions:
More and more data science packages start to drop support for Python 3.7.

For example, `scikit-learn` and `mlflow` no longer support it (PRs #335 and #355)

PR #357, opened 2022-11-29
Scala cat to num unsupervised

cshekhar17:


ionicsolutions:
@cshekhar17 @VishnuGowthem This branch should probably re-based on the current `main` prior to merging, it contains a lot of commits from previously merged branches.

src/main/anovos/data_transformer/transformers.py
```diff
@@ -40,11 +40,6 @@
 from packaging import version
 from scipy import stats
 
-if version.parse(pyspark.__version__) < version.parse("3.0.0"):
-    from pyspark.ml.feature import OneHotEncoderEstimator as OneHotEncoder
-else:
-    from pyspark.ml.feature import OneHotEncoder
-
```
ionicsolutions, line 47: As long as Anovos supports Spark 2.4.8, this switch has to remain in the code (See #356 for the plans to drop support for Python 3.7 and Spark 2.4.8)

PR #358, opened 2022-11-30
updated geo_format_conversion filter conditions

ZhuLi99:
This is to fix the bug in geo_format_conversion part inside workflow.py in the case when all geo-columns are passed but centroid, rog were not calculated.


ranjanravish (approved):
I am good with the changes. I tested it and its working fine now.

